PR #19, opened 2022-08-23
Print received Data

kimhanbeom:
- Modify send data form (test)
- Parse receive data
- Print record type/timestamp

Resolves: #4


kimhanbeom:
해당 이슈 수정 완료 했습니다.

msk (approved):

PR #182, opened 2022-11-18
modify cert issuer cn to subject cn

MW-Kim:
Change the data source from the issuer's cn to the subject's cn.
The subject's cn should determine the certificate's host.


src/server.rs
```diff
@@ -54,7 +54,7 @@ pub fn certificate_info(connection: &Connection) -> Result<String> {
         bail!("invalid X.509 certificate");
     };
     let issuer = x509
```
msk, line 56: 변수 이름도 같이 바꿔주세요.

msk (changes requested):

msk (approved):

PR #289, opened 2023-03-03
Modify database options

BLYKIM:
- Add cf options.
- Modify db open method to enable cf options.
- Add cf options to config file

todo:
- additional options
- docs of config options

Issue: #280

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -13,7 +13,7 @@ use giganto_client::ingest::{
     Packet,
 };
 pub use rocksdb::Direction;
-use rocksdb::{ColumnFamily, DBIteratorWithThreadMode, Options, DB};
+use rocksdb::{ColumnFamily, ColumnFamilyDescriptor, DBIteratorWithThreadMode, Options, DB};
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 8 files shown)


syncpark (approved):

syncpark (approved):

src/storage.rs
```diff
@@ -38,26 +38,50 @@ const RAW_DATA_COLUMN_FAMILY_NAMES: [&str; 14] = [
 const META_DATA_COLUMN_FAMILY_NAMES: [&str; 1] = ["sources"];
 const TIMESTAMP_SIZE: usize = 8;
 
+pub struct DBOptions {
```
msk, line 41: [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case)에 따르면 `DbOptions`로 해야 합니다.

> In `UpperCamelCase`, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`.

msk (approved):

Issue #287, opened 2023-03-01
Modify the GraphQL API, which returns packets, so that it should return parsed content as well


sehkone:
Even though piglet parses some sorts of packets, it does not do so for entire content of them. Plus, when giganto returns packets, they are composed of the packet which piglet already parsed, and some others piglet didn't. Therefore, it would be better to send fully parsed content of packets to users rather packets themselves or with content parsed by piglet itself.

By using tshark for parsing packets, we need to add parsed result in the API returning packets. Since Prof. Park thankfully sent to me a sample code which shows how to run tshark and get parsed data from it, anyone can refer to it.

```
use pcap::{Capture, Linktype, Precision};
use std::{io::Write, os::fd::AsRawFd, path::Path, process::Stdio};

fn main() {
    let args = std::env::args().collect::<Vec<_>>();

    let mut tshark_command = std::process::Command::new("tshark");
    let tshark = tshark_command.arg("-V");
    let mut tshark = match args.len() {
        2 => tshark,
        3 => match args[2].as_str() {
            "ek" | "fields" | "json" | "pdml" => tshark.arg("-T").arg(args[2].as_str()),
            _ => {
                println!("Unknown output format {}", args[1].as_str());
                return;
            }
        },
        _ => {
            println!(
                "Usage: {} <pcap file> [ek|fields|json|pdml]",
                Path::new(args[0].as_str())
                    .file_name()
                    .unwrap()
                    .to_str()
                    .unwrap()
            );
            return;
        }
    }
    .arg("-r")
    .arg("-")
    .stdin(Stdio::piped())
    .stdout(Stdio::piped())
    .spawn()
    .unwrap();

    let mut read_cap =
        Capture::from_file_with_precision(args[1].as_str(), Precision::Nano).unwrap();
    let tshark_stdin = tshark.stdin.take().unwrap();
    let mut write_cap = unsafe {
        Capture::dead_with_precision(Linktype::ETHERNET, Precision::Nano)
            .unwrap()
            .savefile_raw_fd(tshark_stdin.as_raw_fd())
            .unwrap()
    };

    while let Ok(packet) = read_cap.next_packet() {
        write_cap.write(&packet);
    }

    drop(tshark_stdin);

    let output = tshark.wait_with_output().unwrap();
    std::io::stdout().write_all(&output.stdout).unwrap();
}

```

syncpark:
Rust sample code to make PCAP file from packet data(`Vec<u8>`)
```rust
[...]

PR #299, opened 2023-03-12
Mario/add tcpdump

marioEinsis:
Resolves: #287

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,10 @@ jobs:
       with:
         rust-version: stable
     - uses: actions/checkout@v2
+    - name: Install libcap on Ubuntu
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+             sudo apt-get install libpcap-dev libpcap0.8
     - name: Build
       run: cargo build --verbose
     - name: Run tests
@@ -47,6 +51,9 @@ jobs:
       with:
         rust-version: stable
     - uses: actions/checkout@v2
+    - name: Install libcap on Ubuntu
+      run: |
+            sudo apt-get install libpcap-dev libpcap0.8
     - name: Install Tarpaulin
       run: cargo install cargo-tarpaulin
     - name: Generate coverage report
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,9 @@ futures-util = "0.3"
 giganto-client = { git = "https://github.com/aicers/giganto-client.git", tag = "0.1.0" }
 humantime = "2.1"
 humantime-serde = "1"
+libc = "0.2"
 num_enum = "0.5"
+pcap = "1"
 quinn = "0.9"
 rocksdb = "0.20"
 roxy = { git = "https://github.com/aicers/roxy.git", tag = "0.2.0" }
@@ -26,6 +28,7 @@ rustls-pemfile = "1.0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 toml_edit = "0.19"
+tempfile = "3"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 tracing = "0.1"
 warp = { version = "0.3", features = ["tls"] }
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -5,6 +5,7 @@ mod packet;
 pub mod status;
 mod timeseries;
 
+use self::network::NetworkFilter;
 use crate::{
     ingest::{implement::EventFilter, PacketSources},
     storage::{
@@ -19,11 +20,23 @@ use async_graphql::{
 };
 use base64::{engine::general_purpose::STANDARD as base64_engine, Engine};
 use chrono::{DateTime, TimeZone, Utc};
+use giganto_client::ingest::Packet as pk;
+use libc::timeval;
+use pcap::{Capture, Linktype, Packet, PacketHeader};
 use serde::{de::DeserializeOwned, Serialize};
-use std::{net::IpAddr, path::PathBuf, sync::Arc};
+#[cfg(target_os = "macos")]
+use std::os::fd::AsRawFd;
+#[cfg(target_os = "linux")]
+use std::os::unix::io::AsRawFd;
+use std::{
+    io::{Read, Seek, SeekFrom, Write},
+    net::IpAddr,
+    path::PathBuf,
+    process::{Command, Stdio},
+    sync::Arc,
+};
+use tempfile::tempfile;
 use tokio::sync::Notify;
-
-use self::network::NetworkFilter;
 
 pub const TIMESTAMP_SIZE: usize = 8;
 
@@ -85,6 +98,7 @@ pub fn schema(
 /// provided.
 /// Maximum size: 100.
 const MAXIMUM_PAGE_SIZE: usize = 100;
+const A_BILLION: i64 = 1_000_000_000;
 
 fn get_connection<T>(
     store: &RawEventStore<'_, T>,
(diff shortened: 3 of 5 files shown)


syncpark:
@marioEinsis If you change few lines of code in `run_tcudump()` function like following, you can get same result in macos too. I tried to get better result from Tshark, but failed. Currently Tcpdump returns better result.
```rust
    if let Some(mut child_stdin) = child.stdin.take() {
        #[cfg(target_os = "macos")]
        child_stdin.write_all(&[0, 0, 0, 0])?;
        child_stdin.write_all(pcap)?;
        // std::thread::spawn(move || child_stdin.write(pcap));
    } else {
        return Err(anyhow!("failed to execute tcpdump"));
    }
```

syncpark (changes requested):

src/graphql.rs
```diff
@@ -329,6 +360,63 @@ where
...
+            data: &packet.packet,
+        };
+        file.write(&p);
+    }
+    let mut buf = Vec::new();
+    let _ = file.flush();
+    let _ = temp_file.seek(SeekFrom::Start(0));
+    let _ = temp_file.read_to_end(&mut buf)?;
+
+    let cmd = "tcpdump";
```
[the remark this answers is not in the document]
marioEinsis: I tried this last time and the GitHub action workflow produced "" which is why I changed it. However, let me change it and see what happens. Maybe there was typo

src/graphql/packet.rs
```diff
@@ -92,6 +99,41 @@ impl PacketQuery {
...
+        key_prefix.push(0);
+
+        let (start, end) = filter.time();
+
+        let iter = store.boundary_iter(
+            &lower_closed_bound_key(&key_prefix, start),
+            &upper_open_bound_key(&key_prefix, end),
+            Direction::Forward,
+        );
+        let (records, _) = collect_records(iter, 1000, &filter)?;
```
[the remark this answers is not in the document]
marioEinsis: I'm honestly not too sure since a section of this code was made by Bly. But I can make a conditional block

syncpark (approved):

Cargo.toml
```diff
@@ -17,14 +17,17 @@ futures-util = "0.3"
 giganto-client = { path ="giganto-client" }
 humantime = "2.1"
 humantime-serde = "1"
+libc = "0.2.140"
```
sehkone, line 20: Any reason to use the specific version?

src/graphql.rs
```diff
@@ -329,6 +341,61 @@ where
...
+    } else {
+        return Err(anyhow!("failed to execute tcpdump"));
+    }
+
+    let output = child.wait_with_output()?;
+    if output.status.success() {
+        return Ok(String::from_utf8_lossy(&output.stdout).into_owned());
+    }
+
+    Err(anyhow!("failed to run tcpdump"))
```
sehkone, line 397: I think the final return looks better to be an `Ok` rather than `Err` even though I am not sure if doing so generates a clippy warning.

src/graphql.rs
```diff
@@ -19,13 +20,23 @@ use async_graphql::{
...
+use std::os::unix::io::AsRawFd;
+use std::{
+    io::{Read, Seek, SeekFrom, Write},
+    net::IpAddr,
+    path::PathBuf,
+    process::{Command, Stdio},
+};
+use tempfile::tempfile;
 
 pub const TIMESTAMP_SIZE: usize = 8;
```
sehkone, line 39: Before and after a struct definition, insert an empty line, please

src/graphql.rs
```diff
@@ -329,6 +341,61 @@ where
...
+        return Err(anyhow!("failed to execute tcpdump"));
+    }
+
+    let output = child.wait_with_output()?;
+    if output.status.success() {
+        return Ok(String::from_utf8_lossy(&output.stdout).into_owned());
+    }
+
+    Err(anyhow!("failed to run tcpdump"))
+}
```
sehkone, line 398: Insert an empty line as well, please.

src/graphql/packet.rs
```diff
@@ -92,6 +100,44 @@ impl PacketQuery {
...
+
+        if !records.is_empty() {
+            for (_, packet) in records {
+                packet_vector.push(packet);
+            }
+        }
```
sehkone, line 129: Why don't you use like the below instead?

`let packet_vector = records.into_iter().map(|(_, packet)| packet).collect();`

src/graphql/packet.rs
```diff
@@ -51,6 +53,12 @@ struct Packet {
     packet: String,
 }
 
+#[derive(SimpleObject, Debug)]
+struct Pcap {
+    request_time: DateTime<Utc>,
+    parsed_pcap: Vec<String>,
```
sehkone, line 59: Why is this a vector? In your code, this looks having just one item all the time.

src/graphql/packet.rs
```diff
@@ -92,6 +100,44 @@ impl PacketQuery {
...
+        if !records.is_empty() {
+            for (_, packet) in records {
+                packet_vector.push(packet);
+            }
+        }
+
+        match write_run_tcpdump(&packet_vector) {
+            Ok(content) => pcap.push(content),
+            Err(e) => error!("Error parsin pcap : {}", e),
+        };
```
sehkone, line 134: The below is better.

`let pcap = write_run_tcpdump(&packet_vector)?;`

src/graphql/packet.rs
```diff
@@ -92,6 +100,44 @@ impl PacketQuery {
...
+    async fn pcap<'ctx>(&self, ctx: &Context<'ctx>, filter: PacketFilter) -> Result<Pcap> {
+        let db = ctx.data::<Database>()?;
+        let store = db.packet_store()?;
+        let mut packet_vector = Vec::new();
+        let mut pcap = Vec::new();
+        let mut key_prefix = Vec::with_capacity(filter.source.len() + TIMESTAMP_SIZE + 2);
+        key_prefix.extend_from_slice(filter.source.as_bytes());
+        key_prefix.push(0);
+        key_prefix.extend_from_slice(&filter.request_time.timestamp_nanos().to_be_bytes());
+        key_prefix.push(0);
```
sehkone, line 114: This part is mostly the same as in `packets`. Please, remove the repetition.

Cargo.toml
marioEinsis: Yes. This is the most recent version and I just followed the line here : https://crates.io/crates/libc

src/graphql/packet.rs
marioEinsis: When we first started, the code in which Bly provided was Vec<String> at first ,so just in case, I did not want to change it since I wasn't sure if it would break something.

src/graphql/packet.rs
marioEinsis: For packet it's used with : 
```        query(
            after,
            before,
            first,
            last,
            |after, before, first, last| async move {
                load_connection(&store, &key_prefix, &filter, after, before, first, last)
            },
        )
        .await
```
And for Pcap:

```
        let iter = store.boundary_iter(
            &lower_closed_bound_key(&key_prefix, start),
            &upper_open_bound_key(&key_prefix, end),
            Direction::Forward,
        );
        let (records, _) = collect_records(iter, 1000, &filter)?;
```

I can make a function for key_prefix to remove repetition but I'm not sure about "db" and "store" since both those variables are used differently. How can I remove the repetition without changing both functions too much?

Cargo.toml
sehkone: In order to follow any updates, we just have to specify 0.2 without 140

src/graphql.rs
marioEinsis: Done

src/graphql.rs
marioEinsis: Done

src/graphql.rs
marioEinsis: Done

src/graphql/packet.rs
marioEinsis: Done

src/graphql/packet.rs
marioEinsis: Done

src/graphql/packet.rs
marioEinsis: Done

src/graphql/packet.rs
marioEinsis: Done

marioEinsis: Added a function called get_key_prefix_packets

sehkone (approved):

sehkone:
@marioEinsis I think you need to rebase this.

.github/workflows/ci.yml
```diff
@@ -47,9 +51,12 @@ jobs:
...
+    - name: Install libcap on Ubuntu
+      run: |
+            sudo apt-get install libpcap-dev libpcap0.8
     - name: Install Tarpaulin
       run: cargo install cargo-tarpaulin
     - name: Generate coverage report
       run: cargo tarpaulin --out Xml
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v3
```
msk, line 62: The last byte of each line should be a newline character. See https://stackoverflow.com/a/729795.

marioEinsis: Got it! Added a newline after 62

msk: Now it has an extra newline character. The original line was

```
"uses: codecov/codecov-action@v3\n"
```

and you first changed it to

```
"uses: codecov/codecov-action@v3"
```

where I pointed out a newline character is missing at the end.

After the most recent change of yours, it became

```
"uses: codecov/codecov-action@v3\n\n"
```

Please remove the superfluous newline character.

marioEinsis: Done, I think when I saved it in visual studio it automatically adds a new line to the file. So when I added the new line, it added another new line automatically.

marioEinsis: Should have 63 lines instead of 64 lines now

msk (approved):

msk:
Please resolve the conflicts with main.

Issue #331, opened 2023-03-29
Giganto gets list of event Ids from clusters or outliers that matches search filter


marioEinsis:
Aice-web send a vector of event Ids with a search filter as arguments and returns a list of event Ids that matches filter condition. Does not discriminate if outliers or clusters so only API.

sehkone:
Giganto receives a list of event IDs in the format of` Vec<String, Vec<String>>`. The first `String` represents the cluster ID, and the last `String` represents the list of event IDs associated with that cluster. When Giganto returns the result, it should follow the same format.

PR #357, opened 2023-04-19
Add search API

marioEinsis:
Resolves: #331 

Make API to return HTTP event depending on a filter that matches including keyword. If there is something that matches the filter, just return those keys. Else return all keys back as raw events.

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -2,6 +2,7 @@ mod export;
 mod log;
 pub mod network;
 mod packet;
+pub mod search;
 mod source;
 pub mod status;
 mod timeseries;
@@ -43,6 +44,7 @@ pub const TIMESTAMP_SIZE: usize = 8;
 
 #[derive(Default, MergedObject)]
 pub struct Query(
+    search::SearchQuery,
     log::LogQuery,
     network::NetworkQuery,
     export::ExportQuery,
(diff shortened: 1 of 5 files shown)


src/graphql/search.rs
```diff
@@ -0,0 +1,294 @@
...
+        let end = if let Some(end) = &src_ip_range.end {
+            item.orig_addr <= end.parse::<IpAddr>()?
+        } else {
+            false
+        };
+
+        if end || start {
+            return Ok(false);
+        };
+    }
```
sehkone, line 220: I think the following would be clearer

```
if let Some(src_ip_range) = &filter.src_ip {
        let (start, end) =
            if let (Some(start), Some(end)) = (&src_ip_range.start, &src_ip_range.end) {
                (start.parse::<IpAddr>()?, end.parse::<IpAddr>()?)
            } else {
                return Ok(false);
            };

        if item.orig_addr >= start || item.orig_addr <= end {
            return Ok(false);
        }
    }
```

And, I think this part can be a function, so that the below redundant codes will be simpler.

src/graphql.rs
```diff
@@ -2,6 +2,7 @@ mod export;
 mod log;
 pub mod network;
```
sehkone, line 3: I think `pub(super)` would be a better choice.

sehkone:
The design has been changed as we know, I close this PR.

PR #593, opened 2023-11-10
Change ip range and port range to be close ends search

marioCluml:
If a user inputs a single IP or port then search those events only


kimhanbeom (commented):
The existing functionality is currently being used by another project, so I'll check and let you know.

kimhanbeom (approved):

Issue #726, opened 2024-04-23
Update dependencies for mitigating vulnerability


sehkone:
Refer to https://github.com/aicers/giganto/security/dependabot/22

Even `cargo update` might not fail in resolving the above, depending on the dependencies of the crates. If this is the case, please notify us about which crate causes it, and we will discuss it later.

kimhanbeom:
The current dependency for `h2` is
- `warp 0.3.6`/`mockito 1.4.0` -> `hyper 0.14.28` -> `h2 0.3.25`

I checked the repositories for each crate and found that `mockito` has an updated version of `hyper`, but no release for that version yet. and `warp`, there is an issue and pr for `hyper`, but it doesn't seem to be progressing.

sehkone:
> The current dependency for `h2` is
> 
> * `warp 0.3.6`/`mockito 1.4.0` -> `hyper 0.14.28` -> `h2 0.3.25`
> 
> I checked the repositories for each crate and found that `mockito` has an updated version of `hyper`, but no release for that version yet. and `warp`, there is an issue and pr for `hyper`, but it doesn't seem to be progressing.

mockito의 경우 hyper 새 버전 사용이 main 브랜치에는 업데이트가 되었는데 그걸 아직 릴리스하기 전이라는 말인가요?

sehkone:
Could you visit the GitHub pages for Warp and Mockito and set up monitoring to alert you about any status changes?

sehkone:
@kimhanbeom 
This should belong to the "pending" category in our to-do list project, which is the number 10 of our policy.

kimhanbeom:
Since `H2` doesn't require a new release to version up (`0.3.26`), I'll continue with the current issue.

Issue #721, opened 2024-04-22
Update `rustls` to latest version 


sophie-cluml:
- background : https://github.com/aicers/giganto/security/dependabot/23

sophie-cluml:
pending: https://github.com/aicers/patio/issues/29#issuecomment-2073879733

kimhanbeom:
This security vulnerability can be patched via a version update to rustls `0.21.11`, so I'll proceed with the PR (see: https://github.com/aicers/patio/issues/29#issuecomment-2105138712)

PR #727, opened 2024-04-24
Release 0.20.0

kimhanbeom:
related: #717


sehkone:
@sophie-cluml @kimhanbeom , I think it would be better to include #726 and #721 in a new version. Security concerns should be handled with high priority.

kimhanbeom:
@sophie-cluml 
The vulnerability issue has been resolved and I will proceed with the current PR.

Issue #781, opened 2024-07-17
Remove unnecessary dependency


sophie-cluml:
Just found out that `humantime` library is not _directly_ used in our codebase. So I think we need to remove it.

sophie-cluml:
Closed by #782

PR #782, opened 2024-07-17
Remove humantime

Salanoid:
Fix for: Remove unnecessary dependency #781

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


sophie-cluml:
Hello Salanoid, thank you for your interest and contribution in this project. In the current codebase, humantime-serde is directly used, so I think we can remove humantime, but keep humantime-serde.

Salanoid:
Thank you for the reply, I have added the humantime-serde back.

sophie-cluml:
Could you check the failing "CI / check"? I think it can be resolved by updating Cargo.lock file, which I believe can be achieved by `cargo check` or `cargo build`.

sophie-cluml:
It seems like `cargo update` was ran, and it caused update of dependency crates overall. Since we would like to focus on removing humantime for this issue, could you limit the scope of change to humantime?

sophie-cluml (approved):

PR #792, opened 2024-08-05
Fix misleading log 

### LINKED ISSUE #789: Fix misleading log in `certificate_info`
When giganto first starts up itself, it produces below log, even though giganto is not a client to giganto.

```
Connected client name : <agent-name>@<issuer-name>
```

image

This needs improvement, because we do not want to see this log when giganto starts up, but only when clients to giganto connect to giganto.

henry0715-dev:
### Desc

modify funtion - certificate_info_default
add param : print_client_name

### Refenece

Close #789

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -70,6 +70,13 @@ pub fn extract_cert_from_conn(connection: &Connection) -> Result<Vec<Certificate
 }
 
 pub fn certificate_info(cert_info: &[CertificateDer]) -> Result<(String, String)> {
+    certificate_info_default(cert_info, true)
+}
+
+pub fn certificate_info_default(
+    cert_info: &[CertificateDer],
+    print_client_name: bool,
+) -> Result<(String, String)> {
     let Some(cert) = cert_info.first() else {
         bail!("no certificate in identity");
     };
@@ -85,7 +92,9 @@ pub fn certificate_info(cert_info: &[CertificateDer]) -> Result<(String, String)
         .and_then(|cn| cn.as_str().ok())
         .context("the subject of the certificate is not valid")?;
     if subject.contains('@') {
-        info!("Connected client name : {subject}");
+        if print_client_name {
+            info!("Connected client name : {subject}");
+        }
         let parsed = subject.split('@').collect::<Vec<&str>>();
 
         Ok((String::from(parsed[0]), String::from(parsed[1])))
--- a/src/main.rs
+++ b/src/main.rs
@@ -38,7 +38,7 @@ use tracing_subscriber::{
 
 use crate::{
     graphql::{status::TEMP_TOML_POST_FIX, NodeName},
-    server::{certificate_info, Certs, SERVER_REBOOT_DELAY},
+    server::{certificate_info_default, Certs, SERVER_REBOOT_DELAY},
     storage::migrate_data_dir,
 };
 
@@ -160,7 +160,7 @@ async fn main() -> Result<()> {
         let ack_transmission_cnt = new_ack_transmission_count(settings.ack_transmission);
 
         let schema = graphql::schema(
-            NodeName(certificate_info(&cert)?.1),
+            NodeName(certificate_info_default(&cert, false)?.1),
             database.clone(),
             pcap_sources.clone(),
             ingest_sources.clone(),
--- a/src/peer.rs
+++ b/src/peer.rs
@@ -33,8 +33,8 @@ use crate::{
         TomlPeers, CONFIG_GRAPHQL_SRV_ADDR, CONFIG_PUBLISH_SRV_ADDR,
     },
     server::{
-        certificate_info, config_client, config_server, extract_cert_from_conn, Certs,
-        SERVER_CONNNECTION_DELAY, SERVER_ENDPOINT_DELAY,
+        certificate_info, certificate_info_default, config_client, config_server,
+        extract_cert_from_conn, Certs, SERVER_CONNNECTION_DELAY, SERVER_ENDPOINT_DELAY,
     },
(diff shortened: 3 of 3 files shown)


henry0715-dev:
같은 작업
https://github.com/aicers/giganto/pull/791

kimhanbeom:
@henry0715-dev 해당 이슈를 해결하기 위한 중간 단계 커밋을 둘 필요 없어 보입니다. 하나의 commit으로 정리해주세요.

kimhanbeom:
owner의 부재로 Deputy를 assignee로 지정하겠습니다.

kimhanbeom:
https://github.com/aicers/giganto/blob/f7e2781a4958d78a67d6ed484b63a7c0ff197ff4/src/peer.rs#L112-L117
`peer::new` 또한  `certificate_info_default(certs.certs.as_slice(), false)` 로 변경 해야 합니다.

kimhanbeom (changes requested):
https://github.com/aicers/giganto/blob/f7e2781a4958d78a67d6ed484b63a7c0ff197ff4/src/peer.rs#L112-L117
`peer::new` 또한  `certificate_info_default(certs.certs.as_slice(), false)` 로 변경 해야 합니다.

kimhanbeom:
@henry0715-dev 수정 사항에 대한 중간 단계 커밋 없이 하나의 커밋으로 정리해주세요.

henry0715-dev:
@kimhanbeom 
commit 완료하였습니다.
감사합니다.

kimhanbeom (approved):

src/server.rs
```diff
@@ -70,6 +70,13 @@ pub fn extract_cert_from_conn(connection: &Connection) -> Result<Vec<Certificate
 }
 
 pub fn certificate_info(cert_info: &[CertificateDer]) -> Result<(String, String)> {
+    certificate_info_default(cert_info, true)
+}
+
+pub fn certificate_info_default(
```
sehkone, line 76: I don't get what the postfix `_default` exactly mean. @kimhanbeom or @henry0715-dev, could you anyone tell me about it?

henry0715-dev: @sehkone 
함수 명을 certificate_info_print_client_nm 이렇게 수정하면 조금더 직관적일까요?

henry0715-dev: trait을 활용하여 변경하는 것도 고려해 보도록 하겠습니다.

sehkone: @henry0715-dev @kimhanbeom 

I suggest providing two public methods: `subject_from_cert` and `subject_from_cert_verbose`. Additionally, we need to define a private method, such as `subject_from_cert_opt`, as follows:

```Rust
pub fn subject_from_cert(cert_info: &[CertificateDer]) -> Result<(String, String)> {
    subject_from_cert_opt(cert_info, false)
}

pub fn subject_from_cert_verbose(cert_info: &[CertificateDer]) -> Result<(String, String)> {
    subject_from_cert_opt(cert_info, true)
}

fn subject_from_cert_opt(cert_info: &[CertificateDer], logging: bool) -> Result<(String, String)> {
     ...
}
```

The rationale behind this is that we can provide clearer names for users of the public methods, while the function that is difficult to name can remain private. BTW, the word default is hardly meaningful in this context.

henry0715-dev: 최초의 의도는 flag외에도 더 많은 파라미터들이 추가될 수도 있을것이란 가정으로 생각하고 간단하게 만들었습니다.
위에 언급해주신 것과 같은 형태로 만들게 된다면 파라미터 추가 되었을때, 영향이 있는 소스를 수정하는것은 불편함이 발생할 것이라 생각 됩니다.

henry0715-dev: subject_from_cert_opt 메소드명은 좋은것 같습니다.
아래와 같이 2개의 함수로만 사용하는것은 어떨까요?

```
pub fn subject_from_cert_opt(cert_info: &[CertificateDer], logging: bool) -> Result<(String, String)>
pub fn subject_from_cert(cert_info: &[CertificateDer]) -> Result<(String, String)>
```

sehkone: @henry0715-dev 
While your suggestion is reasonable, it’s not certain that we will need additional options besides this logging in the future. Since it would be safe to follow your suggestion if more options become necessary, I believe the focus should be on providing public methods with clearer names for now. I think the prefix **opt** is too vague, so I believe it is more appropriate to use it in private methods, in this context.

sehkone: What I wanted to point out is that `certificate_info` only calls another public method, `certificate_info_default`. According to your suggestion, `subject_from_cert` would call `subject_from_cert_opt` in the same way, even though `subject_from_cert_opt` can do exactly what `subject_from_cert` does. In other words, `subject_from_cert_opt` provides two different features, but one method provides only one of these features, while no other method provides the other, which creates an imbalance. In this context, the two mutually exclusive features should be provided separately: one with logging and the other without logging. From the user's perspective, it would be easier to understand if the two different methods each provided only one of these two features.

henry0715-dev: @sehkone 
새로운 이슈 생성해서 말씀주신 아래의 코드로 변경하도록 하겠습니다.
```
pub fn subject_from_cert(cert_info: &[CertificateDer]) -> Result<(String, String)> {
    subject_from_cert_opt(cert_info, false)
}

pub fn subject_from_cert_verbose(cert_info: &[CertificateDer]) -> Result<(String, String)> {
    subject_from_cert_opt(cert_info, true)
}

fn subject_from_cert_opt(cert_info: &[CertificateDer], logging: bool) -> Result<(String, String)> {
     ...
}
```

Issue #823, opened 2024-09-10
Release 0.21.0


kimhanbeom:
## Checklist

- [x] Submit PR
- [x] Tag annotated with version number
- [x] Push tags

kimhanbeom:
pending: https://github.com/aicers/giganto-client/issues/132

sophie-cluml:
Releasing Giganto needs to address these issues as well:
- #815
- #804

sophie-cluml:
I think both #820 #819 can be resolved shortly, and it is related to node management project, so I think we can wait just a little bit.

sophie-cluml:
Tagged : https://github.com/aicers/giganto/releases/tag/0.21.0

PR #837, opened 2024-09-23
Release 0.21.0

### LINKED ISSUE #836: Combine `migrate_0_19_to_0_21_0_alpha_1` and `migrate_0_21_0_alpha_1_to_0_21_0_alpha_2`
While preparing the release of Giganto, I noticed that two migration functions, `migrate_0_19_to_0_21_0_alpha_1` and `migrate_0_21_0_alpha_1_to_0_21_0_alpha_2`, can be combined into a single migration without any side effects.

- `migrate_0_19_to_0_21_0_alpha_1` converts the `Conn` struct from [this version](https://github.com/aicers/giganto/blob/main/src/storage/migration/migration_structures.rs#L40-L52) to [this one](https://github.com/aicers/giganto/blob/main/src/storage/migration/migration_structures.rs#L55-L68).
- `migrate_0_21_0_alpha_1_to_0_21_0_alpha_2` then converts `Conn` from [this version](https://github.com/aicers/giganto/blob/main/src/storage/migration/migration_structures.rs#L55-L68) to the latest `Conn` structure in `giganto-client`.

Since both migrations target the `Conn` data up to Giganto version 0.21.0, combining them will eliminate the need for multiple migration steps, significantly reducing the time spent migrating `Conn` data, which constitutes the majority of our dataset.

sophie-cluml:
related to #823
close #836 

This PR is related to two issues, since #836 and #823 cannot be easily split into separate PRs, mainly because #836 could be considered as a subset task of #823.

### DIFF
--- a/src/storage/migration.rs
+++ b/src/storage/migration.rs
@@ -14,8 +14,7 @@ use serde::de::DeserializeOwned;
 use tracing::info;
 
 use self::migration_structures::{
-    ConnBeforeV21A1, ConnFromV21A1BeforeV21A2, HttpFromV12BeforeV21, NtlmBeforeV21, SmtpBeforeV21,
-    SshBeforeV21, TlsBeforeV21,
+    ConnBeforeV21, HttpFromV12BeforeV21, NtlmBeforeV21, SmtpBeforeV21, SshBeforeV21, TlsBeforeV21,
 };
 use super::Database;
 use crate::{
@@ -27,7 +26,7 @@ use crate::{
     },
 };
 
-const COMPATIBLE_VERSION_REQ: &str = ">=0.21.0-alpha.2,<0.22.0";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.21.0,<0.22.0";
 
 /// Migrates the data directory to the up-to-date format if necessary.
 ///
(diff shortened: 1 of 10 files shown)


msk:
@sophie-cluml, thanks for working on the new release of Giganto (and the migration optimization—that was a good catch!). I noticed that this PR is using a version of `giganto-client` that's a few commits behind the latest release (0.20.0). Since @petabi plans to migrate its applications to `giganto-client` 0.20.0, I'm concerned about potential compatibility issues with Giganto 0.21.0. Would it be possible to update the `giganto-client` dependency in this PR to the latest release to ensure compatibility?

kimhanbeom:
https://github.com/aicers/giganto/blob/100798454e55fc4c77304167962e883fac136113/src/storage/migration.rs#L436-L438

Could you please change it to `migrate_0_19_to_0_21_0`?

src/storage/migration.rs
```diff
@@ -613,13 +593,13 @@ mod tests {
         tls_store.append(&key, &ser_old_tls).unwrap();
 
         //migration 0.19.0 to 0.21.0-alpha.1
```
kimhanbeom, line 595:
```suggestion
        //migration 0.19.0 to 0.21.0
```

kimhanbeom (changes requested):

src/storage/migration.rs
sophie-cluml: Done

kimhanbeom (approved):

Issue #849, opened 2024-10-03
Update async-graphql to avoid a security issue, and release a new version


sehkone:
Refer to [this](https://github.com/aicers/giganto/security/dependabot/27).

@kimhanbeom,

I noticed that the explicit listing of the task to run `cargo update` was missing from the issue for the last release, as shown below. Even though it might have been done, without including the task in the list, we can't immediately tell if it was actually completed.

<img width="766" alt="Screenshot 2024-10-04 at 8 27 02 AM" src="https://github.com/user-attachments/assets/2800b469-e103-40b7-9c50-8e68373af0d3">

In the last release, I guess that you did `cargo update`, but it was before the release of the async-graphql patch.

@sophie-cluml,

You created the commit message "Release 0.21.0", which I don't think is ideal because the actual release happens only after we tag the new version. Therefore, the commit message should have been something like "Bump version to 0.21.0". Sometimes, after bumping the version, we realize that additional changes need to be made for the release, and we have to push another commit, which is then tagged with the new version. In such cases, the tagged version will point to one of the commits made after the version bump. I remember we had a similar situation in another repository and I said the same argument to you.

<img width="1549" alt="Screenshot 2024-10-04 at 8 36 40 AM" src="https://github.com/user-attachments/assets/2d5802d0-4cf0-4bac-b372-4a5aa2388a5d">

kimhanbeom:
## Checklist

- [ ] Check dependencies.
   - [ ] Check that all dependencies use the release version.
   - [ ] Check that dependencies are up-to-date with `cargo update`.
- [ ] Submit PR
- [ ] Tag annotated with version number
- [ ] Push tags

kimhanbeom:
The issue has been resolved, so I'll close it

PR #848, opened 2024-10-02
Update `config` to include local information

BLYKIM:
Close: #817

PR #850, opened 2024-10-04
Bump version to 0.22.0

kimhanbeom:
- Updated dependency for security vulnerabilities.
- Update the compatibility version of the quic communication modules.

Related issue: #849


sophie-cluml:
I'd like to suggest including https://github.com/aicers/giganto/pull/848 in the new release, which is currently waiting for an additional review from @sehkone. Would it be possible?

sophie-cluml (approved):

sehkone:
> I'd like to suggest including #848 in the new release, which is currently waiting for an additional review from @sehkone. Would it be possible?

I think I missed this request. I am sorry about that. But It's ok to proceed anyway.

Issue #834, opened 2024-09-20
Remove logs that print configuration


sophie-cluml:
Some configurations are not meant to be exposed to users, as they are intended to remain hidden. Since users have access to the logs and often view log files, it is safer to remove logs that print configuration details. This will help protect sensitive information from being inadvertently exposed.

Related GraphQL API is `setConfig`.

PR #851, opened 2024-10-08
Remove configuration details from info macro

marioCluml:
Closes: #834 ;


marioCluml:
I removed the part where in `info!` it prints the config details. Also checked any `warn!`, `error!` and `debug!`. However, I was wondering if there are any other that comes to mind? I can't seem to find anything else at the moment.

Issue #879, opened 2024-10-29
Suggestion: Restrict DB Repair Functionality to Local Configuration Mode Only


sophie-cluml:
Currently, the `repair` command-line option in Giganto is intended for manually triggered database repair processes, typically used by engineers to resolve issues in maintenance situations. Since this option is not meant for routine operation, I propose limiting the DB repair functionality to only work in local configuration mode.

This means, that if the `repair` option is specified without the `-c` option to provide a local configuration, Giganto should terminate, as proceeding without local configuration is not supported for repair operations.

PR #883, opened 2024-11-04
Modify the repair mode

BLYKIM:
Close: #879

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,7 +92,11 @@ async fn main() -> Result<()> {
         settings.config.num_of_thread,
         settings.config.max_sub_compactions,
     );
+
     if args.repair {
+        if !is_local_config {
+            bail!("repair is not allowed on remote config");
+        }
         let start = Instant::now();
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -90,7 +90,8 @@ async fn main() -> Result<()> {
         settings.config.num_of_thread,
         settings.config.max_sub_compactions,
     );
-    if args.repair {
+
+    if is_local_config && args.repair {
```
sophie-cluml, line 94: I think this if conditional can be simplified for humans like

```
if args.repair {
  if !is_local_config {
   // bail
  } else {
    // do repair
  }
}
```

This structure might make it easier to follow the logic at a glance. Only if you agree, I'd like to suggest changing like above.

sehkone: I think Clippy would say that what @sophie-cluml suggested doesn’t follow one of its lints.

sophie-cluml: @sehkone My suggestion does not produce Clippy warning.

sophie-cluml (approved):

Issue #885, opened 2024-11-07
Return error when `setConfig` GraphQL API is called when Giganto is running with local conf


sophie-cluml:
It is now returning `Ok(false)`. We need to return an error.

PR #886, opened 2024-11-07
Update `setConfig` GraphQL API

sophie-cluml:
Closes: #885

### DIFF
--- a/src/graphql/status.rs
+++ b/src/graphql/status.rs
@@ -180,8 +180,8 @@ impl ConfigMutation {
         let is_local = ctx.data::<bool>()?;
 
         if *is_local {
-            warn!("Config is local");
-            return Ok(false);
+            warn!("The request to the GraphQL API to change the configuration was ignored because this is running with a local configuration.");
+            return Err("Config is local".to_string().into());
         }
 
         let config_draft: Config = toml::from_str(&draft)?;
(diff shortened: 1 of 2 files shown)


src/graphql/status.rs
```diff
@@ -181,7 +181,7 @@ impl ConfigMutation {
 
         if *is_local {
             warn!("Config is local");
-            return Ok(false);
+            return Err("Config is local".to_string().into());
```
sehkone, line 184: The `Err` message is sufficiently clear to API users, but the warning message may not be as clear to those who see it among log messages. In fact, users attempting to set the config remotely is more significant than simply running with a local configuration. Since the API users themselves triggered this action, they don't need to be notified of it. However, those reviewing the log messages should be notified, as they may lack this context.

sophie-cluml: Elaboration is added.

src/graphql/status.rs
```diff
@@ -180,8 +180,8 @@ impl ConfigMutation {
         let is_local = ctx.data::<bool>()?;
 
         if *is_local {
-            warn!("Config is local");
-            return Ok(false);
+            warn!("Configuration set GraphQL API request disregarded: Giganto is running with a local configuration, so the request to set the configuration is disregarded.");
```
sehkone, line 183: I suggest the following to clarify the message:

The request to the GraphQL API to change the configuration was ignored because this is running with a local configuration.

sophie-cluml: Done

sehkone (approved):

Issue #896, opened 2024-11-15
Add GraphQL API to provide raw events in csv format


kimhanbeom:
Whenever a new protocol is added, we have to modify the code to display it in the UI, and to minimize this modification,  we need to add a GraphQL API that provides the raw events of network/sysmon protocols as a csv-formatted string.
(See UI repository issue 745 for more details)

PR #897, opened 2024-11-18
Add GraphQL API `csvFormattedRawEvents`

kimhanbeom:
- Add GraphQL API `csvFormattedRawEvents` that returns the values of raw events of the request protocol in csv format String.

Close: #896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added GraphQL API `csvFormattedRawEvents` that returns the values of raw
+  events of the request protocol in csv format String, delimited by tab.
+
 ### Changed
 
 - Remote configuration is no longer stored in a temporary file, nor does it
@@ -22,7 +27,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
     family. Running this version of Giganto will migrate the existing data in
     `sources` column familiy to `sensors` column family.
   - The `sources` GraphQL API is renamed to `sensors`.
-  - The `sourceId` field in the `export` GraphQL API  is renamed to `sensorId`.
+  - The `sourceId` field in the `export` GraphQL API is renamed to `sensorId`.
   - The `source` field in the filter parameters of all GraphQL APIs is changed
     to `sensor`.
 
--- a/src/graphql/log.rs
+++ b/src/graphql/log.rs
@@ -1,7 +1,10 @@
 #[cfg(test)]
 mod tests;
 
-use std::{fmt::Debug, net::IpAddr};
+use std::{
+    fmt::{Debug, Display},
+    net::IpAddr,
+};
 
 use anyhow::anyhow;
 use async_graphql::{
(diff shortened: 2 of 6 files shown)


syncpark (approved):

sehkone:
@kimhanbeom @syncpark 

As I posted in the number 71 of our discussions, we need to restrict the term "log" to mean only the log message that our software generate. The external data must be not referred the term "log" to avoid confusion.

src/graphql/log.rs
```diff
@@ -1,33 +1,58 @@
...
 
 #[derive(Default)]
 pub(super) struct LogQuery;
 
+#[allow(clippy::module_name_repetitions)]
+#[derive(InputObject, Serialize)]
+struct CsvFilter {
+    protocol: String,
+    timestamps: Vec<DateTime<Utc>>,
+    source: String,
```
sophie-cluml, line 53: I think this can be sensor instead.

kimhanbeom: sensor is correct, so I fixed it.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added GraphQL API `csvFormattedRawEvents` that returns the values of raw
+  events of the request protocol in csv format String.
```
sophie-cluml, line 13: I think it would be worth describing that the delimiter is tab, not comma, such as "the request protocol in csv format string, delimited by tab." Other than this point, I think this PR is good to be merged.

sophie-cluml (approved):

sehkone:
@sophie-cluml @kimhanbeom 

Why would you use tab instead of comma?

kimhanbeom:
@sehkone 
In the current UI, we are outputting raw events (network, sysmon) using tab as a delimiter (using the display trait). The return value of `csvFormattedRawEvents` should also be in the same format as output by the existing UI, so I used the same delimiter(tab) as in the UI.

sehkone:
> @sehkone In the current UI, we are outputting raw events (network, sysmon) using tab as a delimiter (using the display trait). The return value of `csvFormattedRawEvents` should also be in the same format as output by the existing UI, so I used the same delimiter(tab) as in the UI.

Does this mean we don’t have a specific reason to keep it this way? If not, I think we should switch to using a comma as the delimiter, as it’s the most common method and aligns with the name CSV.

kimhanbeom:
> > @sehkone In the current UI, we are outputting raw events (network, sysmon) using tab as a delimiter (using the display trait). The return value of `csvFormattedRawEvents` should also be in the same format as output by the existing UI, so I used the same delimiter(tab) as in the UI.
> 
> Does this mean we don’t have a specific reason to keep it this way? If not, I think we should switch to using a comma as the delimiter, as it’s the most common method and aligns with the name CSV.

Here's an additional reason to use tab as a delimiter

In giganto-client, we implemented a display trait for each raw event, and it uses the same tab as the UI as a delimiter.
That's why using tabs as delimiters makes it easier to implement `csvFormattedRawEvents` and doesn't require special modifications if new protocols are added in the future.

but If we use a separate delimiter instead of tab, then it would be extra work to implement a separate trait in the structure of each raw event to generate the result of `csvFormattedRawEvents`, and that would lead to more work in giganto due to the API added to “minimize modifications in the UI repository every time a new protocol is added”, so it seems better to keep the delimiter as tab.

sehkone:
I don't understand why CSV-formatted strings should depend on the implementation of `Display`. Is there any specific reason to keep using it to convert events to CSV? On the contrary, I think we have two options:

- Modify the implementation of `Display` to use a comma.
- Leave it as is, but add dedicated code to generate CSV strings. (I prefer this)

What I want to point out is that there is no inherent dependency between how `Display` is implemented and the format of a CSV, yet you seem to be assuming such a dependency exists.

kimhanbeom:
I spoke with @sehkone about the `csvFormattedRawEvents` and here's what we talked about.

- `csvFormattedRawEvents` is an API that was added for the purpose of taking tab-delimited results from the UI and outputting them to the screen itself. 
- However, in general, users will see the name `csvFormattedRawEvents` and assume that the result of the API call will be in csv (comma-separated values) format.
- So I'll rename the API to `tsvFormattedRawEvents` to make it clear that we're using tab delimiters.

Issue #895, opened 2024-11-14
Remove deleted `source` within the `Netflow5`, `Netflow9`, `SecuLog`


kimhanbeom:
Related issue: https://github.com/aicers/giganto-client/issues/139

Issue #877, opened 2024-10-29
Rename `source` to `sensor`


kimhanbeom:
- Change `source` to `sensor`, which is a more appropriate term for the name of the device that sensed/captured the raw event.

PR #898, opened 2024-11-18
Rename source to sensor

sophie-cluml:
Fixes: #877 
Fixes: #895 

This PR handles renaming mainly, and follow up changes that is necessary after renaming.

- Below are directly related to renaming, and are fairly straightforward:
  - The `source` field in the filter parameters of all GraphQL APIs is changed
    to `sensor`.
  - The `sourceId` field in the `export` GraphQL API  is renamed to `sensorId`.
  - The `sources` GraphQL API is renamed to `sensors`.

- Below are related to the renaming, but since it might be hard to navigate through the changed code, so I'd like to elaborate:
  - Migrated `sources` cf to `sensors` cf. This involves copying the data from the old column family to the new one, and then dropping the old column family.
  - Fixed GraphQL schema for Giganto cluster. This was necessary for 2 reasons. One is because renaming a field cause a change in GraphQL API interface, and Giganto cluster is affected by the GraphQL API change. The other is because the graphql.schema file was not correctly reflecting recent changes, and it seemed like the file was manually modified. So I felt it is necessary to fix it, by replacing the file to use the automatically generated schema.graphql from the GraphQL playground.
  - ~Removed `source` field from the `netflow5RawEvents`, `netflow9RawEvents`, and `secuLogRawEvents` GraphQL APIs. As a result, `Netflow5`, `Netflow9`, and `SecuLog` no longer use `source` (now renamed to `sensor`) as their database start key. Instead, they use an empty string for the start key. This approach allows us to focus solely on the renaming without restructuring the database storage keys at this point. I think the restructuring of database keys is a broader topic that requires a more comprehensive discussion, which we can address separately. For now, I think using an empty string as the start key respects the current setup while accommodating the necessary changes.~ This striked-out sentences should be disregarded. Instead, `source` field is removed form `Netflow5`, `Netflow9` and `SecuLog` structs. This removal is worked toward the conclusion mentioned in below comment https://github.com/aicers/giganto/pull/898#issuecomment-2482277460.


sophie-cluml:
@kimhanbeom AFAIK, I believe you were also interested in tackling #877 and #895, So for this part,

> Removed source field from the netflow5RawEvents, netflow9RawEvents, and secuLogRawEvents GraphQL APIs. As a result, Netflow5, Netflow9, and SecuLog no longer use source (now renamed to sensor) as their database start key. Instead, they use an empty string for the start key. This approach allows us to focus solely on the renaming without restructuring the database storage keys at this point. I think the restructuring of database keys is a broader topic that requires a more comprehensive discussion, which we can address separately. For now, I think using an empty string as the start key respects the current setup while accommodating the necessary changes.

I would appreciate if you could share your thoughts on this matter. Do you have other suggestion on how to handle the start key of the `Netflow5`, `Netflow9`, and `SecuLog` events?

sophie-cluml:
In an online meeting with @kimhanbeom, we discussed that issue #895 was specifically focused on removing the `source` field from the event struct. This does not necessarily imply that the `source` should be removed from the DB keys for those structs. Based on this discussion, I will make the necessary adjustments to this draft PR and mark it as ready for review once the changes are complete.

src/storage/migration/migration_structures.rs
```diff
@@ -284,3 +287,202 @@ impl From<TlsBeforeV21> for TlsFromV21 {
...
+    }
+    fn log_contents(&self) -> Option<String> {
+        Some(self.contents.clone())
+    }
+    fn sensor(&self) -> Option<String> {
+        Some(self.source.clone())
+    }
+}
+
+impl From<Netflow9BeforeV23> for crate::storage::Netflow9 {
```
kimhanbeom, line 417: It would be nice to unify how structure names are used within `migration_structures`. 
(rename to alias or use existing structure name).

CHANGELOG.md
```diff
@@ -16,12 +16,26 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
+    to `sensor`.
 
 ### Removed
 
 - Removed the GraphQL API `setAckTransmissionCount` as the entire configuration
   is now sent at once when modified through the UI.
 
+### Fixed
+
+- Fixed a missing update to the schema.graphql file necessary for
```
kimhanbeom, line 36:
```suggestion
- Fixed a missing update to the schema. graphql file necessary for
```

src/storage/migration.rs
```diff
@@ -8,40 +8,46 @@ use std::{
 };
 
 use anyhow::{anyhow, Context, Result};
-use giganto_client::ingest::log::SecuLog;
+use giganto_client::ingest::{
```
kimhanbeom, line 11: It would be preferable to modify the code to access it using crate::storage instead of referencing the structure directly from giganto-client in order to clearly indicate the purpose of changing the DB schema. (ex)`crate::storage::Netflow5`)

CHANGELOG.md
sophie-cluml: It means the file schema.graphql, so I think it should not have space.

src/storage/migration/migration_structures.rs
sophie-cluml: Sounds good, done.

src/storage/migration.rs
sophie-cluml: Done

kimhanbeom (approved):
LGTM

Issue #917, opened 2024-12-03
Add Documentation Comments for GraphQL APIs - `Conn` Related GraphQL APIs


sophie-cluml:
For the parent issue https://github.com/aicers/giganto/issues/887, I would like to work on `Conn` related GraphQL APIs, which are `connRawEvents` and `networkRawEvents`.

PR #918, opened 2024-12-03
Replace timestamp with time & Add document for GraphQL API `connRawEvents`

sophie-cluml:
Close: #917
Close: #922 

-------------------
Below was a description in PR draft stage:

Hi @sehkone, I’ve created this draft PR to ask for your confirmation if I’m heading in the right direction. So far, I’ve only worked on the `connRawEvents` GraphQL API as a single case example. If the approach aligns with your expectations, I can proceed to apply the same methodology to the remaining GraphQL APIs. Let me know your thoughts.

GraphQL Playground looks like below:

image


src/graphql/network.rs
```diff
@@ -126,23 +126,57 @@ impl RawEventFilter for SearchFilter {
     }
 }
 
+/// Represents an event extracted from a session.
 #[derive(SimpleObject, Debug, ConvertGraphQLEdgesNode)]
 #[graphql_client_type(names = [conn_raw_events::ConnRawEventsConnRawEventsEdgesNode, network_raw_events::NetworkRawEventsNetworkRawEventsEdgesNodeOnConnRawEvent])]
 struct ConnRawEvent {
+    /// TODO: Waiting for an answer in the general repository.
     timestamp: DateTime<Utc>,
+    /// Source IP, or source IP address in long format.
```
sehkone, line 135: `Source IP address ` only would suffice without `Source IP`

src/graphql/network.rs
```diff
@@ -126,23 +126,57 @@ impl RawEventFilter for SearchFilter {
...
     orig_addr: String,
+    /// Source port, or source port number in long format.
```
sehkone, line 137: `Source port number` would suffice.

src/graphql/network.rs
sehkone: What does `long format` means?

sophie-cluml: It is now modified to `Source IP address.`. Long format was what I got from consultation with ChatGPT. I don't think we need it anymore.

src/graphql/network.rs
sophie-cluml: Done

src/graphql/network/tests.rs
```diff
@@ -3217,7 +3217,7 @@ async fn union() {
...
...01T00:00:05+00:00\", __typename: \"SmtpRawEvent\"}}, {node: {time: \"2020-01-01T00:01:01+00:00\", __typename: \"ConnRawEvent\"}}, {node: {timestamp: \"2020-01-05T00:01:01+00:00\", __typename: \"RdpRawEvent\"}}, {node: {timestamp: \"2020-01-05T06:05:00+00:00\", __typename: \"DceRpcRawEvent\"}}, {node: {timestamp: \"2020-06-01T00:01:01+00:00\", __typename: \"HttpRawEvent\"}}, {node: {timestamp: \"2021-01-01T00:01:01+00:00\", __typename: \"DnsRawEvent\"}}, {node: {timestamp: \"2022-01-05T00:01:01+00:00\", __typename: \"NtlmRawEvent\"}}, {node: {timestamp: \"2023-01-05T00:01:01+00:00\", __typename: \"KerberosRawEvent\"}}, {node: {timestamp: \"2023-01-05T12:12:00+00:00\", __typename: \"FtpRawEvent\"}}, {node: {timestamp: \"2023-01-05T12:12:00+00:00\", __typename: \"MqttRawEvent\"}}, {node: {timestamp: \"2023-01-06T11:11:00+00:00\", __typename: \"TlsRawEvent\"}}, {node: {timestamp: \"2023-01-06T12:12:00+00:00\", __typename: \"LdapRawEvent\"}}, {node: {timestamp: \"2023-01-06T12:12:10+00:00\", __typename: \"SmbRawEvent\"}}, {node: {timestamp: \"2023-01-06T12:13:00+00:00\", __typename: \"NfsRawEvent\"}}, {node: {timestamp: \"2023-01-06T12:13:10+00:00\", __typename: \"DhcpRawEvent\"}}]}}");
```
sehkone, line 3220: `timestamp` here does not need to change to `time`?

sophie-cluml: `{node: {timestamp: \"2020-01-01T00:01:01+00:00\", __typename: \"ConnRawEvent\"}}`

is changed to

`{node: {time: \"2020-01-01T00:01:01+00:00\", __typename: \"ConnRawEvent\"}}`

The change of `timestamp` to `time` occurs here partially only for `ConnRawEvent`, not for other events. It is because `networkRawEvents` is a GraphQL query that returns an [Union type](https://graphql.org/learn/schema/#union-types), and in this case `ConnRawEvent` is one of the types included in that Union type response. Since this PR targets to change `ConnRawEvent`, only above part of code is modified.

sophie-cluml: By the way, do you think it is better to modify `timestamp` to `time` in all event protocols all at once, in this PR?

sehkone: Once we decided to changed `timestamp` to `time`, I think it would be better to complete it all at once.

sophie-cluml: I understood. I created an issue regarding timestamp-to-time as #922, and this PR has been updated to resolve that issue together at once.

sophie-cluml:
@sehkone 시간 괜찮으시다면 리뷰 부탁드릴 수 있을까요? 최근 push 들은 단순 rebase들 입니다.

CHANGELOG.md
```diff
@@ -35,6 +35,13 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   - If the configuration is received successfully via the `setConfig` GraphQL
     API, it will switch to normal mode, where all task and DB migrations run
     as normal, just like when running Giganto with local settings.
+- The term timestamp is replaced with time in event structs where the type is
```
sehkone, line 38: By using backtick, `timestamp` and `time` would be better

src/graphql/client/schema/schema.graphql
sehkone, line 1: I was just wondering how this file is regularly updated. Could you share some details on it?

CHANGELOG.md
sophie-cluml: Done. There were `timestamp` and `timestamps`, I changed it to enumerate both cases, to use backticks.

src/graphql/client/schema/schema.graphql
sophie-cluml: The principle is that, whenever GraphQL interface changes, the schema.graphql file changes.
The SDL file used here is acquired from the GraphQL playground.

sehkone (approved):

Issue #899, opened 2024-11-20
Remove `unwrap`


dayeon5470:
## Background
- The development rules for our product say that we should use `expect` instead of `unwrap` in situations where we are confident that `panic` will not necessarily occur.

## Task
- Remove unwrap(), which is used against the rules

sophie-cluml:
@dayeon5470 We should not mention private links.

PR #937, opened 2024-12-23
Refactor unwrap to improve error handling

BLYKIM:
Close: #899

### DIFF
--- a/src/graphql/export.rs
+++ b/src/graphql/export.rs
@@ -2669,8 +2669,9 @@ fn export_statistic_file(
             loop {
                 match iter.next() {
                     Some(Ok(item)) => {
-                        // replace new value (min_index's vector value is always exist)
-                        *iter_next_values.get_mut(min_index).unwrap() = item;
+                        *iter_next_values
+                            .get_mut(min_index)
+                            .expect("The index is one of the actual elements in the vector, so it is always valid.") = item;
                         break;
                     }
                     Some(Err(_)) => {
--- a/src/graphql/statistics.rs
+++ b/src/graphql/statistics.rs
@@ -224,8 +224,7 @@ fn gen_statistics(
         for idx in next_candidate {
             if let Some(iter) = stats_iters.get_mut(idx) {
                 if let Some(item) = iter.next() {
-                    // replace new value (value is always exist)
-                    *iter_next_values.get_mut(idx).unwrap() = item;
+                    *iter_next_values.get_mut(idx).expect("`next_candidate` is generated during iteration over `iter_next_values`, ensuring all its indices are valid within the latter.") = item;
                 } else {
                     // No value to call with the iterator.
                     let _ = stats_iters.remove(idx);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -165,8 +165,10 @@ extern crate proc_macro;
 /// ```
 #[proc_macro_derive(ConvertGraphQLEdgesNode, attributes(graphql_client_type))]
 pub fn derive_from_graphql_client_autogen(input: TokenStream) -> TokenStream {
-    derive_from_graphql_client_autogen_2(input.into())
-        .unwrap()
+    derive_from_graphql_client_autogen_2(input.clone().into())
+        .unwrap_or_else(|e| {
+            panic!("ConvertGraphQLEdgesNode macro is not correctly used for {input:?}: {e}")
+        })
         .into()
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,10 @@ use std::{
     fs,
     net::SocketAddr,
     path::Path,
-    sync::{Arc, Mutex},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
     time::Duration,
 };
 
@@ -126,7 +129,7 @@ async fn main() -> Result<()> {
         let notify_shutdown = Arc::new(Notify::new());
         let notify_reboot = Arc::new(Notify::new());
         let notify_power_off = Arc::new(Notify::new());
-        let retain_flag = Arc::new(Mutex::new(false));
+        let retain_flag = Arc::new(AtomicBool::new(false));
 
         let database = if let Some(settings) = settings.clone() {
             let (db_path, db_options) = db_path_and_option(
@@ -312,11 +315,8 @@ async fn main() -> Result<()> {
 
         if is_reboot || is_power_off || is_reload_config {
             loop {
-                {
-                    let retain_flag = retain_flag.lock().unwrap();
-                    if !*retain_flag {
-                        break;
-                    }
+                if !retain_flag.load(Ordering::Relaxed) {
+                    break;
                 }
                 sleep(Duration::from_millis(SERVER_REBOOT_DELAY)).await;
             }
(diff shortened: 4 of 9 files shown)


src/lib.rs
```diff
@@ -166,7 +166,7 @@ extern crate proc_macro;
 #[proc_macro_derive(ConvertGraphQLEdgesNode, attributes(graphql_client_type))]
 pub fn derive_from_graphql_client_autogen(input: TokenStream) -> TokenStream {
     derive_from_graphql_client_autogen_2(input.into())
-        .unwrap()
+        .expect("valid input struct")
```
sophie-cluml, line 169: Could you handle this with `panic!` macro, instead of `expect` please? The file lib.rs is the code for proc-macro , which runs against the Giganto codebase, during compile time only, to generate code from our custom macro. And this part can panic, when the Giganto codebase uses the custom macro incorrectly. My personal idea is that the message could be something like "ConvertGraphQLEdgesNode macro is not correctly used for {input}"

src/publish/implement.rs
```diff
@@ -13,14 +13,16 @@ pub trait RequestStreamMessage {
 
 impl RequestStreamMessage for RequestHogStream {
     fn channel_key(&self, sensor: Option<String>, record_type: &str) -> Result<Vec<String>> {
+        let sensor = sensor
+            .ok_or_else(|| anyhow!("Failed to generate hog channel key, sensor is required."))?;
```
sophie-cluml, line 17: I think we can use supervised model as the term.

sophie-cluml:
Could you check the commit message title's length?

src/graphql/export.rs
```diff
@@ -2669,8 +2669,9 @@ fn export_statistic_file(
             loop {
                 match iter.next() {
                     Some(Ok(item)) => {
-                        // replace new value (min_index's vector value is always exist)
-                        *iter_next_values.get_mut(min_index).unwrap() = item;
+                        *iter_next_values
+                            .get_mut(min_index)
+                            .expect("min_index's vector value always exists") = item;
```
sehkone, line 2674: I don't think this message is clear enough. I suggest:

`"The index is one of the actual elements in the vector, so it is always valid."`

src/graphql/statistics.rs
```diff
@@ -224,8 +224,7 @@ fn gen_statistics(
         for idx in next_candidate {
             if let Some(iter) = stats_iters.get_mut(idx) {
                 if let Some(item) = iter.next() {
-                    // replace new value (value is always exist)
-                    *iter_next_values.get_mut(idx).unwrap() = item;
+                    *iter_next_values.get_mut(idx).expect("value always exists") = item;
```
sehkone, line 227: We need to specify the reason as clearly as possible, for example:

```rust
"`next_candidate` is generated while iterating over `iter_next_value`, so any index of the former is always valid within the latter."
```

src/main.rs
```diff
@@ -313,7 +313,7 @@ async fn main() -> Result<()> {
         if is_reboot || is_power_off || is_reload_config {
             loop {
                 {
-                    let retain_flag = retain_flag.lock().unwrap();
+                    let retain_flag = retain_flag.lock().expect("Mutex lock is safe");
```
sehkone, line 316: Could you explain why this is safe?

src/publish.rs
```diff
@@ -528,8 +528,7 @@ where
             info!("start hog's publish stream : {:?}", record_type);
         }
         NodeType::Crusher => {
-            // crusher's policy Id always exists.
-            let id = msg.id().unwrap();
+            let id = msg.id().expect("crusher's policy Id always exists");
```
sehkone, line 531: Could you explain why this always exists?

src/storage/migration.rs
```diff
@@ -196,7 +196,7 @@ where
             continue;
         };
         let new_key = StorageKey::builder()
-            .start_key(&netflow_raw_event.sensor().unwrap())
+            .start_key(&netflow_raw_event.sensor().expect("always exists"))
```
sehkone, line 199: Could you explain why this is safe?

src/publish.rs
kimhanbeom: - The `id` method's value is a method that gets the value of the id field of the `RequestCrusherStream` structure that cruhser sends to giganto.
- When crusher sends the value of that structure to giganto, it always includes the id value, so the value always exists.

```rust
pub struct RequestCrusherStream {
    pub start: i64,
    pub id: String,
    pub src_ip: Option<IpAddr>,
    pub dst_ip: Option<IpAddr>,
    pub sensor: Option<String>,
}
```

Apart from this issue, the return value was Option because we used to call one method, `source_id`, to get the source or id values from the `RequestStreamMessage` trait, but now we have separated them into `sensor` and `id`.  `Id` always has a value, so I think we can remove Option from the return value of the `id` method of the `RequestStreamMessage` trait in the future.

src/storage/migration.rs
kimhanbeom: - The `sensor` is a method to get the value of the source field of `Netflow5BeforeV23` and `Netflow9BeforeV23`.
- In earlier versions of giganto, before the structure of netflow5 and netflow9 was changed, saving `Netflow5BeforeV23` and `Netflow9BeforeV23` to the DB was done in the following order
  - Receive netflow5, netflow9 from REproduce. (via ingest)
  -   - Extract the source from the certificate in REproduce and store the value in the source field of each structure. 
  - Store the final struct values to the DB.
- Due to the above processing order, the source value always exists, and so do the results of the `sensor` that fetch them.

src/main.rs
kimhanbeom: If the message in expect about why the `retain_flag` is always safe needs to change, please also change the message about the `running_flag` in the `retain_periodically` function as well.

src/main.rs
```diff
@@ -313,7 +313,7 @@ async fn main() -> Result<()> {
         if is_reboot || is_power_off || is_reload_config {
             loop {
                 {
-                    let retain_flag = retain_flag.lock().unwrap();
+                    let retain_flag = retain_flag.lock().expect("Mutex lock is safe because it only guards a simple boolean flag without complex operations.");
```
sehkone, line 316: I don't think this message sufficiently explains why we can always expect safe results.

BLYKIM: Initially, I used the “always safe” message in expect because the Mutex only protects a simple boolean flag, which makes it unlikely to panic within the lock scope.
However, after considering the possibility of other parts of the system panicking while holding the lock, I decided to replace expect with unwrap_or_else.
In case of poisoning, the value is now safely recovered, and a warning message is logged to aid debugging.

src/storage.rs
```diff
@@ -1038,7 +1041,10 @@ pub async fn retain_periodically(
                 }
                 info!("Database cleanup completed.");
                 {
-                    let mut running_flag = running_flag.lock().unwrap();
+                    let mut running_flag = running_flag.lock().unwrap_or_else(|e| {
+                        warn!("Poisoned Mutex. Recovering the boolean flag.");
```
sophie-cluml, line 1045: I think we can use AtomicBool type instead of `running_flag`. I think we can avoid potential panicking from the mutex behavior, and achieve simpler code.

sophie-cluml (approved):

kimhanbeom (approved):

sehkone (approved):

sophie-cluml:
@BLYKIM Could you check the line length of the commit content?

Issue #932, opened 2024-12-19
Fix Clippy errors in tests and enforce Clippy checks in CI


sophie-cluml:
We haven’t checked Clippy lints in test code before, but now it’s time to do so.

PR #942, opened 2024-12-26
Resolve Clippy errors in test code

danbi2990:
Updated CI configuration to include Clippy checks for tests.

Close: #932

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -1834,7 +1834,7 @@ mod tests {
             let ingest_sensors = Arc::new(tokio::sync::RwLock::new(
                 CURRENT_GIGANTO_INGEST_SENSORS
                     .into_iter()
-                    .map(|sensor| sensor.to_string())
+                    .map(str::to_string)
                     .collect::<HashSet<String>>(),
             ));
 
@@ -1846,7 +1846,7 @@ mod tests {
             let ingest_sensors = Arc::new(tokio::sync::RwLock::new(
                 CURRENT_GIGANTO_INGEST_SENSORS
                     .into_iter()
-                    .map(|sensor| sensor.to_string())
+                    .map(str::to_string)
                     .collect::<HashSet<String>>(),
             ));
 
@@ -1855,7 +1855,7 @@ mod tests {
                 PeerInfo {
                     ingest_sensors: PEER_GIGANTO_2_INGEST_SENSORS
                         .into_iter()
-                        .map(|sensor| (sensor.to_string()))
+                        .map(str::to_string)
                         .collect::<HashSet<String>>(),
                     graphql_port: Some(port),
                     publish_port: None,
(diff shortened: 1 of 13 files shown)


src/graphql.rs
```diff
@@ -1834,7 +1834,7 @@ mod tests {
             let ingest_sensors = Arc::new(tokio::sync::RwLock::new(
                 CURRENT_GIGANTO_INGEST_SENSORS
                     .into_iter()
-                    .map(|sensor| sensor.to_string())
+                    .map(&str::to_string)
```
sehkone, line 1837: I don't think `&` is necessary. Function names are in general considered function pointers.

danbi2990: Applied the suggested adjustment. Thank you.

src/graphql/log/tests.rs
```diff
@@ -518,22 +519,23 @@ async fn load_oplog() {
...
-    key_list.push(insert_oplog_raw_event(&store, "review", "src1", 2, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "review", "src1", 3, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "piglet", "src1", 3, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "piglet", "src1", 4, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "piglet", "src2", 5, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "piglet", "src2", 6, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "review", "src1", 4, &generator).await);
-    key_list.push(insert_oplog_raw_event(&store, "review", "src2", 5, &generator).await);
+    let key_list: Vec<Vec<u8>> = vec![
+        insert_oplog_raw_event(&store, "piglet", "src1", 1, &generator),
```
sophie-cluml, line 528: I think we need to use the public names of the the private repositories in these lines. I Could you modify these please? Besides, I think this could even be done in a separate issue if there are a lot of lines to rename.

danbi2990: I think it’s better to address them in a separate issue, but I have a question: there are quite a few semi-supervised engine names—most of them in identifiers, with a minority in string literals. Is renaming the identifiers required?
In other words, what's the scope of the renaming?

sophie-cluml: I only meant the sensor and the manager modules only in the above comment, but if you want to extend the scope to the semi supervise engine, I agree that a separate issue will be nicer, and the identifiers I believe should be renamed along with the literals. And it would also require updating giganto-client. Let me create the issues in bothd repos.

sophie-cluml (approved):

src/graphql/log/tests.rs
danbi2990: Understood. I will address the sensor and manager names in this PR and handle the semi-supervised engine names in the separate issues.

Issue #955, opened 2025-01-09
Use `if let` or `let else` instead of `unwrap_or_else`, if necessary


danbi2990:
https://github.com/aicers/giganto/blob/b096c59d5856541da084970325a271ff5267fa84/src/lib.rs#L205

https://github.com/aicers/giganto/blob/b096c59d5856541da084970325a271ff5267fa84/src/lib.rs#L211

https://github.com/aicers/giganto/blob/b096c59d5856541da084970325a271ff5267fa84/src/main.rs#L199-L200

Since `unwrap_or_else` implies returning an alternative value to avoid a panic, we should use `if let` or `let else` when actions like printing messages or handling a panic are involved.

PR #961, opened 2025-01-15
Refactor `unwrap_or_else`

BLYKIM:
- Refactor unwrap_or_else to explicitly handle errors instead of returning default values.

Close: #955

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -165,11 +165,10 @@ extern crate proc_macro;
 /// ```
 #[proc_macro_derive(ConvertGraphQLEdgesNode, attributes(graphql_client_type))]
 pub fn derive_from_graphql_client_autogen(input: TokenStream) -> TokenStream {
-    derive_from_graphql_client_autogen_2(input.clone().into())
-        .unwrap_or_else(|e| {
-            panic!("ConvertGraphQLEdgesNode macro is not correctly used for {input:?}: {e}")
-        })
-        .into()
+    match derive_from_graphql_client_autogen_2(input.clone().into()) {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -165,11 +165,10 @@ extern crate proc_macro;
...
 pub fn derive_from_graphql_client_autogen(input: TokenStream) -> TokenStream {
-    derive_from_graphql_client_autogen_2(input.clone().into())
-        .unwrap_or_else(|e| {
-            panic!("ConvertGraphQLEdgesNode macro is not correctly used for {input:?}: {e}")
-        })
-        .into()
+    match derive_from_graphql_client_autogen_2(input.clone().into()) {
+        Ok(result) => result.into(),
+        Err(e) => panic!("ConvertGraphQLEdgesNode macro is not correctly used for {input:?}: {e}"),
+    }
```
sehkone, line 171: I think `let else` would be better in this context.

BLYKIM:
```rust
let Ok(result) = derive_from_graphql_client_autogen_2(input.clone().into()) else {
    panic!("ConvertGraphQLEdgesNode macro is not correctly used for {input:?}");
};
result.into()
```

If the code is changed to use let-else as shown above, it won’t allow access to the error in Err(e). This could result in a loss of detail in the log. Would this be acceptable?

sehkone: I am sorry. I missed that the `e` is already being handled. The original should remain unchanged.

sophie-cluml (approved):

Issue #889, opened 2024-11-07
Create GraphQL API to Indicate Giganto's Idle Mode Status


sophie-cluml:
To allow the Manager to recognize when Giganto is operating in ~minimal~ idle mode, we can implement a new GraphQL API endpoint that returns the current mode status of Giganto. This API will indicate whether Giganto is in ~minimal~ idle mode, enabling the Manager to handle such scenarios appropriately.

sophie-cluml:
Due to the direction explained in https://github.com/aicers/giganto/issues/1015, this PR needs to be closed without merge. This issue is contrary to the newly created issue.

PR #963, opened 2025-01-15
Add `idleMode` GraphQL API

sophie-cluml:
Closes: #889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   it as an optional parameter. Additionally, the API response now includes logs
   from all agents displayed in chronological order, rather than being limited to
   the logs of a single agent.
+- Added the `idleMode` GraphQL API to check if Giganto is in idle mode.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


src/graphql.rs
```diff
@@ -152,6 +152,7 @@ pub struct NodeName(pub String);
 pub struct RebootNotify(Arc<Notify>); // reboot
 pub struct PowerOffNotify(Arc<Notify>); // shutdown
 pub struct TerminateNotify(Arc<Notify>); // stop
+pub(crate) struct MinimalMode(bool);
```
sehkone, line 155: The phrase, minimal mode, was not strictly defined but just expressive, so requires further explanation to convey its meaning to others. Back then, Dr. Kim must haven't pick any stricter term due to the reason that the scope could not be yet well defined. However, since we have developed what it should do in the perspective of daemon behavior, I think we can choose an alternative term that can mean truer and narrower. I suggest "Idle Mode".

sophie-cluml: Thank you for very kind suggestion. Changes are done.

sehkone: The commit message hasn't been modified accordingly.

sophie-cluml: Sorry I missed it. It is modified now.

CHANGELOG.md
```diff
@@ -26,15 +27,15 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
+  - If local configuration is not used when running Giganto, it will run in idle
+    mode, with only the GraphQL server running. Retain task, peer server task,
+    ingest task, publish task, and DB migration are not executed.
+  - In idle mode, the APIs provided by the GraphQL server are limited. In this
+    mode, only the APIs provided by `graphql::status` are available.
   - If the configuration is received successfully via the `setConfig` GraphQL
-    API, it will switch to normal mode, where all task and DB migrations run
-    as normal, just like when running Giganto with local settings.
+    API, it will switch to normal mode, where all task and DB migrations run as
+    normal, just like when running Giganto with local settings.
```
sehkone, line 38: I suggest the following:

```markdown
- Added idle mode to improve the process of setting configuration via the
  GraphQL API `setConfig`.
  - If Giganto starts without a local configuration, it enters idle mode, 
    where it only responds to the GraphQL APIs provided by `graphql::status`, including `setConfig`,
    while excluding all other operations such as retaining, peer server communication, and DB migration.
  - If Giganto in idle mode successfully receives a configuration via `setConfig`,
    it exits idle mode and functions normally as if it had started with a local configuration.
```

src/graphql/client/schema/schema.graphql
```diff
@@ -1652,6 +1652,10 @@ type Query {
   propertiesCf(filter: PropertyFilter!): Properties!
   config: Config!
   ping: Boolean!
+
+  # Checks if the server is in idel mode. In this mode, Giganto is not fully operational and
```
sehkone, line 1656: A typo, `idel`

src/graphql/status.rs
```diff
@@ -183,6 +183,13 @@ impl StatusQuery {
     async fn ping(&self) -> Result<bool> {
         Ok(true)
     }
+
+    /// Checks if the server is in idel mode. In this mode, Giganto is not fully operational and
```
sehkone, line 187: The same typo

CHANGELOG.md
sehkone: I think there is one thing unclear in what I suggested above. How does Giganto behave if it starts without a local configuration but has already received a configuration earlier?

sophie-cluml: > I think there is one thing unclear in what I suggested above. How does Giganto behave if it starts without a local configuration but has already received a configuration earlier?

I believe this case is explained by what you suggested:

``
- If Giganto in idle mode successfully receives a configuration via `setConfig`, it exits idle mode and functions normally as if it had started with a local configuration.
```

sehkone:
@sophie-cluml,

According to our phone discussion, I think we need to save the received configuration so that it can be used when Giganto restarts. Could you create another issue for this?

sehkone (approved):

kimhanbeom:
@sehkone 
https://github.com/aicers/giganto/blob/187c3b1863afe2af0d2366ee106f08aec0f61460/src/main.rs#L280-L290
Currently, Giganto is implemented to save remote configuration when they are received via `write_toml_file`.

sophie-cluml:
Due to the direction explained in https://github.com/aicers/giganto/issues/1015, this PR needs to be closed without merge.

Issue #964, opened 2025-01-16
Add Documentation Comments for GraphQL APIs - `Dns` Related GraphQL APIs


danbi2990:
For the parent issue #887, I would like to work on `Dns` related GraphQL APIs.

PR #984, opened 2025-01-21
Add document for `DnsRawEvent`

danbi2990:
Close: #964


src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
+    /// Query Class. Range: 1 - 4.
+    ///
+    /// [Domain Name System Parameters.](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4)
     qclass: u16,
+    /// Query Type.
+    ///
+    /// - 1: Internet Class
+    /// - 2: CSnet Class
+    /// - 3: Chaos Class
+    /// - 4: HeSiod Class
```
sophie-cluml, line 220: From the https://www.ietf.org/rfc/rfc1035.html#section-3.2.4, I think the correct class names are
CSNET and CHAOS and Hesiod classes. And with these names I think we don't need to add the clippy suppression `#[allow(clippy::doc_markdown)]`.

src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
     last_time: StringNumberI64,
+    /// Query.
     query: String,
+    /// Answer.
     answer: Vec<String>,
+    /// Transaction ID.
     trans_id: u16,
+    /// Round-Trip Time.
     rtt: StringNumberI64,
+    /// Query Class. Range: 1 - 4.
```
sophie-cluml, line 211: This is a trivial suggestion.

```suggestion
    /// Query class. Range: 1-4.
    /// Query class. Range is from 1 to 4.
```

Reason behind is that I think we usually don't put spaces in between to express number ranges. (I just consulted chatgpt, because I think chatgpt writes better English than me.) But chatgpt and I might be wrong. So please ignore this comment if 1 - 4 is correct.

src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
     orig_addr: String,
+    /// Source port number.
     orig_port: u16,
+    /// Destination IP address.
     resp_addr: String,
+    /// Destination port number.
     resp_port: u16,
+    /// Protocol number. TCP is 6, UDP is 17.
     proto: u8,
+    /// End Time. time + rtt (Default).
```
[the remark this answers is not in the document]
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
     query: String,
+    /// Answer.
     answer: Vec<String>,
+    /// Transaction ID.
     trans_id: u16,
+    /// Round-Trip Time.
     rtt: StringNumberI64,
+    /// Query Class. Range: 1 - 4.
+    ///
+    /// [Domain Name System Parameters.](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4)
```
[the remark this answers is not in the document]
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
     proto: u8,
+    /// End Time. time + rtt (Default).
     last_time: StringNumberI64,
+    /// Query.
     query: String,
+    /// Answer.
     answer: Vec<String>,
+    /// Transaction ID.
     trans_id: u16,
+    /// Round-Trip Time.
```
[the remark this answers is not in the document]
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
     tc_flag: bool,
+    /// Recursion Desired Flag.
+    ///
+    /// Recursion Desired flag of DNS query and response message.
     rd_flag: bool,
+    /// Recursion Available Flag.
+    ///
+    /// Only from DNS response message.
     ra_flag: bool,
+    /// Time to live.
```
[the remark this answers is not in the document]
danbi2990: Applied the suggested adjustment.
You're correct unless the Sensor changes the unit.
image

src/graphql/network.rs
```diff
@@ -180,28 +180,90 @@ struct ConnRawEvent {
...
+    /// |16|BADSIG|TSIG Signature Failure|
+    /// |17|BADKEY|Key not recognized|
+    /// |18|BADTIME|Signature out of time window|
+    /// |19|BADMODE|Bad TKEY Mode|
+    /// |20|BADNAME|Duplicate key name|
+    /// |21|BADALG|Algorithm not supported|
+    /// |22|BADTRUNC|Bad Truncation|
+    /// |23|BADCOOKIE|Bad/missing Server Cookie|
     rcode: u16,
+    /// Authoritative Answer Flag.
```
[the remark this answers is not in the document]
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/client/schema/schema.graphql
```diff
@@ -266,25 +266,104 @@ type DnsEventEventEdge {
...
+  #
+  # Truncated flag of DNS query and response message.
   tcFlag: Boolean!
+
+  # Recursion Desired Flag.
+  #
+  # Recursion Desired flag of DNS query and response message.
   rdFlag: Boolean!
+
+  # Recursion Available Flag.
```
sophie-cluml, line 361: What about Flag to flag in these comments as well?

danbi2990: I think it's better to clarify the convention for the remaining tasks.
The initial version I wrote is from the Notion document, but it's preferable to capitalize only the first character, right?

sophie-cluml: Yes you are right.

danbi2990: Applied the suggested adjustment. Thank you.

src/graphql/client/schema/schema.graphql
```diff
@@ -266,25 +266,104 @@ type DnsEventEventEdge {
...
+  # Destination IP address.
   respAddr: String!
+
+  # Destination port number.
   respPort: Int!
+
+  # Protocol number. TCP is 6, UDP is 17.
   proto: Int!
+
+  # End time in nanoseconds. `time + rtt` (Default).
```
sophie-cluml, line 289: Just one last thing. I can't find `time + rtt` in our codebase. Could it be `time` + `rtt` ?

danbi2990: Applied the suggested adjustment.

sophie-cluml (approved):

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
     orig_addr: String,
+    /// Source port number.
     orig_port: u16,
+    /// Destination IP address.
     resp_addr: String,
+    /// Destination port number.
     resp_port: u16,
+    /// Protocol number. TCP is 6, UDP is 17.
     proto: u8,
+    /// End time in nanoseconds. `time` + `rtt` (Default).
```
sehkone, line 200: `(Default)` looks confusing because it implies there might be other cases. I assume this `last_time` is not parsed but calculated from `time` + `rtt` by the software, and `(Default)` doesn't seem to describe it accurately. I also think the content from `time` might not be necessary. @syncpark, could you confirm this?

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
+    /// Query class. Range: 1-4.
+    ///
+    /// For more information, refer to [Domain Name System Parameters.](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4)
     qclass: u16,
+    /// Query type.
+    ///
+    /// - 1: Internet class
+    /// - 2: CSNET class
+    /// - 3: CHAOS class
+    /// - 4: Hesiod class
```
sehkone, line 219: These values belong to `qclass`, not `qtype`. In the document where attributes are described, content other than my suggested titles has not been verified. Therefore, we should not copy it unless it has been strictly verified.

sehkone: @sophie-cluml,

For now, I think it would be better to include just the titles, not the additional content written in the document. If such additional content is included in another context, we should either remove it or strictly verify it.

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
+    /// |12-15|Unassigned||
+    /// |16|BADVERS|Bad OPT Version|
+    /// |16|BADSIG|TSIG Signature Failure|
+    /// |17|BADKEY|Key not recognized|
+    /// |18|BADTIME|Signature out of time window|
+    /// |19|BADMODE|Bad TKEY Mode|
+    /// |20|BADNAME|Duplicate key name|
+    /// |21|BADALG|Algorithm not supported|
+    /// |22|BADTRUNC|Bad Truncation|
+    /// |23|BADCOOKIE|Bad/missing Server Cookie|
```
sehkone, line 247: I think only remaining `Response code` would be better.

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
     rcode: u16,
+    /// Authoritative answer flag.
+    ///
+    /// Only from DNS response message.
```
sehkone, line 251: I think this line is unnecessary.

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
     aa_flag: bool,
+    /// Truncated flag.
+    ///
+    /// Truncated flag of DNS query and response message.
```
sehkone, line 255: So does this.

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
     tc_flag: bool,
+    /// Recursion desired flag.
+    ///
+    /// Recursion desired flag of DNS query and response message.
```
sehkone, line 259: The same

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
     rd_flag: bool,
+    /// Recursion available flag.
+    ///
+    /// Only from DNS response message.
```
sehkone, line 263: This one as well.

src/graphql/network.rs
```diff
@@ -180,28 +180,89 @@ struct ConnRawEvent {
...
     query: String,
+    /// Answer.
     answer: Vec<String>,
+    /// Transaction ID.
     trans_id: u16,
+    /// Round-trip time in nanoseconds.
     rtt: StringNumberI64,
+    /// Query class. Range: 1-4.
+    ///
+    /// For more information, refer to [Domain Name System Parameters.](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4)
```
sehkone, line 212: If we include a reference in one context, we should do the same in every context. Therefore, I think it would be better to remove it.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: Applied the suggested adjustment.

src/graphql/network.rs
danbi2990: @syncpark 
Could you please review this comment? It suggests the following change.
```
- End time in nanoseconds. `time` + `rtt` (Default).
+ End time in nanoseconds.
```

sophie-cluml: @danbi2990 Let's just make this part the same as we did in other events, such as https://github.com/aicers/giganto/pull/1021/files#diff-353689134b6a76113df88a614d4b93d4a96a3bcacf8bc163837255dbec1c2286R720-R722

src/graphql/client/schema/schema.graphql
```diff
@@ -272,25 +272,63 @@ type DnsEventEventEdge {
...
+  # Source port number.
   origPort: Int!
+
+  # Destination IP address.
   respAddr: String!
+
+  # Destination port number.
   respPort: Int!
+
+  # Protocol number. TCP is 6, UDP is 17.
```
sophie-cluml, line 292: I think we can modify this part similarly to other events as well.

danbi2990: Updated the code like the others.

src/graphql/client/schema/schema.graphql
```diff
@@ -389,25 +389,67 @@ type DnsEventEventEdge {
...
+  # Truncated Flag
   tcFlag: Boolean!
+
+  # Recursion Desired Flag
   rdFlag: Boolean!
+
+  # Recursion Available Flag
   raFlag: Boolean!
+
+  # Time to live
```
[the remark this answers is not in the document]
danbi2990: Applied the suggested adjustment.

sophie-cluml (approved):

sophie-cluml:
@sehkone I believe this PR is well aligned with our policy regarding the GraphQL API documentation, since we have already synced in other similar PRs. And I also believe the points that you have raised in previous comments are all resolved. So let me merge this PR. If you think we should not merge, I'd appreciate if you could let us know.

Issue #987, opened 2025-01-24
Update `config` GraphQL API to response in "{n}d" format


sophie-cluml:
Currently, Giganto responds like below:

```
{
  "data": {
    "config": {
      "ingestSrvAddr": "127.0.0.1:38370",
      "publishSrvAddr": "127.0.0.1:38371",
      "graphqlSrvAddr": "127.0.0.1:8442",
      "retention": "3months 8days 16h 19m 12s",
      "dataDir": "/opt/clumit/var/giganto",
      "logDir": "/opt/clumit/log",
      "exportDir": "/opt/clumit/var/giganto/export",
      "maxOpenFiles": 8000,
      "maxMbOfLevelBase": "512",
      "numOfThread": 8,
      "maxSubCompactions": "2",
      "addrToPeers": null,
      "ackTransmission": 1234
    }
  }
```

We would like to update "retention" to return "100d" instead.

PR #988, opened 2025-01-24
Update `retention` format in `config` GraphQL API

sophie-cluml:
Closes: #987


sophie-cluml:
I think that’s a valid point, @kimhanbeom. However, the `config` GraphQL API only returns the configuration when Giganto is running in remote config mode. Since I think your point involves around local config mode, I would like to suggest handling that in a separate issue and a PR. What do you think about this?

kimhanbeom (approved):

Issue #878, opened 2024-10-29
Adjust Logging Behavior by Revised Logging Policy


sophie-cluml:
## Logging Behavior

~### Local Configuration~
~- If `log_dir` is not set, logs should be directed to `stdout`.~
~- If `log_dir` is set but is not writable, the application should terminate and provide a message to the user indicating the issue.~

~### Remote Configuration~
~- If `log_dir` is not set, logs should be directed to `stdout`.~
~- If `log_dir` is set but is not writable, the application should switch to a minimal mode:~
  ~- In this mode, a message should be sent to the server indicating that the `log_dir` is not writable.~
  ~- Any logs generated while in minimal mode should be directed to `stdout`.~

~This structured approach ensures that logging is handled gracefully in both configurations, while providing useful feedback to users or administrators on logging issues.~

~### Notes~

~The function for notifying the server when `log_dir` is not writable—sending a message to the server—can be implemented separately. We may handle this part in a separate issue at https://github.com/aicers/giganto/issues/889.~

The new logging policy is described here: https://github.com/aicers/giganto/issues/878#issuecomment-2626102212.

sophie-cluml:
The discussion regarding the new logging policy has evolved a bit, and it is now like below. Besides, I crossed out all the contents in the main description.

- **For local config:**
  - If `log_dir` is **not present** in the TOML config → Logs are written to stdout/stderr. 
  - If `log_dir` is **present** in the TOML config and writable → Logs are written to the specified `log_dir`.
  - If `log_dir` is **present** in the TOML config but not writable → The program terminates. 

- **For remote config:**
  - If `log_dir` is **not present** in the remote config → Logs are written to stdout/stderr. However, this situation does not occur because `log_dir` is declared as a constant in UI repo. 
  - If `log_dir` is **present** in the remote config and writable → Logs are written to the specified `log_dir`.
  - If `log_dir` is **present** in the remote config but not writable → The program enters idle(minimal) mode. 
 
- Special question: Where to log before fetching the remote config?
  - Logs are written to stdout/stderr until the remote config is retrieved.

PR #993, opened 2025-02-03
Adjust logging behavior by revised logging policy

kimhanbeom:
- If `log_dir` is not present in the local/remote config, logs are written to stdout/stderr.
- If `log_dir` in the local/remote config is writable, logs are written to the specified `log_dir`.
- If `log_dir` in the local config is not writable, the program terminates.
- If `log_dir` in the remote config is not writable, the program enters idle mode.
- Logging in idle mode: logs are written to stdout/stderr until the remote config is retrieved.

Close: #878

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -152,6 +152,7 @@ pub struct NodeName(pub String);
 pub struct RebootNotify(Arc<Notify>); // reboot
 pub struct PowerOffNotify(Arc<Notify>); // shutdown
 pub struct TerminateNotify(Arc<Notify>); // stop
+pub struct TracingEnabled(bool);
 
 #[allow(clippy::too_many_arguments)]
 pub fn schema(
@@ -169,6 +170,7 @@ pub fn schema(
     ack_transmission_cnt: AckTransmissionCount,
     is_local_config: bool,
     settings: Option<Settings>,
+    tracing_enabled: bool,
 ) -> Schema {
     Schema::build(Query::default(), Mutation::default(), EmptySubscription)
         .data(node_name)
@@ -185,6 +187,7 @@ pub fn schema(
         .data(PowerOffNotify(notify_power_off))
         .data(is_local_config)
         .data(settings)
+        .data(TracingEnabled(tracing_enabled))
         .finish()
 }
 
@@ -195,6 +198,7 @@ pub fn minimal_schema(
     notify_terminate: Arc<Notify>,
     is_local_config: bool,
     settings: Option<Settings>,
+    tracing_enabled: bool,
 ) -> MinimalSchema {
     MinimalSchema::build(
         MinimalQuery::default(),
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -479,3 +536,20 @@ pub async fn notify_and_wait_shutdown(is_minimal_mode: bool, notify_shutdown: Ar
         notify_shutdown.notified().await;
     }
 }
+
+fn log(tracing_log_enable: bool, log_type: Level, log_message: &str) {
```
sophie-cluml, line 540: I think the parameter names can be slightly refined for better clarity and readability. Since the function name already makes it clear that it handles logging, we can remove the `log_` part from the parameters. Additionally, `log_type` could be renamed to `level`, as "level" is already a well-understood term in logging contexts. Lastly, changing `tracing_log_enable` to `tracing_enabled` makes it more natural by using an adjective form.  

```rust
fn log(tracing_enabled: bool, level: Level, message: &str)
```  

Let me know what you think.

src/graphql.rs
```diff
@@ -152,6 +152,7 @@ pub struct NodeName(pub String);
 pub struct RebootNotify(Arc<Notify>); // reboot
 pub struct PowerOffNotify(Arc<Notify>); // shutdown
 pub struct TerminateNotify(Arc<Notify>); // stop
+pub struct TracingLogEnable(bool);
```
sophie-cluml, line 155: Perhaps we can go for `TracingEnabled` to align with my suggestion for the `log` function, if you agree.

src/main.rs
```diff
@@ -429,47 +467,66 @@ fn new_peers_data(peers_list: Option<HashSet<PeerIdentity>>) -> (Peers, PeerIden
...
+///
+/// If `log_dir` is `None`, logs will be printed to stdout.
+/// If the runtime is in debug mode, logs will be printed to stdout in addition to the specified
+/// `log_dir`.
+///
+/// # Errors
+///
+/// Returns an error if the log file cannot be opened in the `log_dir` path in the
+/// local configuration.
+pub fn init_tracing(log_dir: Option<&Path>) -> Result<Vec<WorkerGuard>> {
```
sophie-cluml, line 480: Can we remove the `pub` here, if it is unnecessary?

kimhanbeom: done

src/graphql.rs
kimhanbeom: done

src/graphql.rs
```diff
@@ -169,6 +170,7 @@ pub fn schema(
     ack_transmission_cnt: AckTransmissionCount,
     is_local_config: bool,
     settings: Option<Settings>,
+    tracing_log_enable: bool,
```
sophie-cluml, line 173: What do you think about turning the `tracing_log_enable` to `tracing_enabled` for consistent variable naming?

kimhanbeom: done

sophie-cluml (approved):

Issue #994, opened 2025-02-04
Change the `log_dir` type in `Config` to `Option<PathBuf>`


kimhanbeom:
The revised logging policy requires logs to be written to stdout if `log_dir` does not present, so we need to change the `log_dir` type in `Config` to `Option<PathBuf>` to allow log_dir to be optionally set.

PR #995, opened 2025-02-04
Change the `log_dir` type to `Option<PathBuf>`

kimhanbeom:
Close: #994

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   - `ftpRawEvents`
 - Changed `config` GraphQL API to respond `retention` field in "{days}d" format
   to align with the format of the configuration field in the API request.
+- `log_dir` is not a required configuration item.
 
 ### Removed
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ giganto -c <CONFIG_PATH> --cert <CERT_PATH> --key <KEY_PATH> --ca-certs \
 ### Arguments
 
 - `<CONFIG_PATH>`: Path to the TOML configuration file (optional when running in
-    remote mode).
+  remote mode).
 - `<CERT_PATH>`: Path to the certificate file (required).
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   - `networkRawEvents`
 - Changed `config` GraphQL API to respond `retention` field in "{days}d" format
   to align with the format of the configuration field in the API request.
+- Changed the `log_dir` field type in `Config` structure to `Option<PathBuf>`.
```
sophie-cluml, line 48: Since CHANGELOG's main audience is users rather than contributors, I think this can be modified in the interest of the main audience, and I suggest something like:

- `log_dir` is not a required configuration item(or entry).

Besides, I think it would be beneficial to tell what happens when `log_dir` is not present at all in the configuration, so that users understand what happens when `log_dir` is not present. (Maybe this part is going to be covered in the other PR #993. Is that what you have in mind? If that is the case, I assume that this CHANGELOG entry will be modified or removed in that PR, and I think that is fine.)

sophie-cluml, line 48: I think we also need to update README.

kimhanbeom: done

kimhanbeom: done

README.md
```diff
@@ -56,8 +56,8 @@ peers = [ { addr = "10.10.12.1:38383", hostname = "ai" } ]     # list of peer in
 
 By default, giganto reads the config file from the following directories:
 
-* Linux: `$HOME/.config/giganto/config.toml`
-* macOS: `$HOME/Library/Application Support/com.cluml.giganto/config.toml`
+- Linux: `$HOME/.config/giganto/config.toml`
+- macOS: `$HOME/Library/Application Support/com.cluml.giganto/config.toml`
```
sophie-cluml, line 60: I think this is not up-to-date. Could you try rebasing?

kimhanbeom: done

sophie-cluml (approved):

PR #1009, opened 2025-02-05
Add documentation for `SmbRawEvent`

### LINKED ISSUE #973: Add Documentation Comments for GraphQL APIs - `Smb` Related GraphQL APIs
For the parent issue #887, I would like to work on `Smb` related GraphQL APIs.

danbi2990:
Closes #973

### DIFF
--- a/src/graphql/client/schema/schema.graphql
+++ b/src/graphql/client/schema/schema.graphql
@@ -537,7 +537,7 @@ type FtpRawEvent {
   # Data channel destination port number.
   dataRespPort: Int!
 
-  # File name.
+  # Filename.
   file: String!
 
   # File size.
@@ -745,10 +745,10 @@ type KerberosRawEvent {
   # End time in nanoseconds.
   lastTime: StringNumberI64!
 
-  # Client time in nanoseconds.
+  # Client time.
   clientTime: StringNumberI64!
 
-  # Server time in nanoseconds.
+  # Server time.
   serverTime: StringNumberI64!
 
   # Error code.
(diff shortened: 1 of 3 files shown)


src/graphql/client/schema/schema.graphql
```diff
@@ -745,10 +745,10 @@ type KerberosRawEvent {
   # End time in nanoseconds.
   lastTime: StringNumberI64!
 
-  # Client time in nanoseconds.
+  # Client time.
```
danbi2990, line 748: It appears that the values are given in nanoseconds as follows.
https://github.com/aicers/piglet/blob/fc7d716bea7197d5ec206913ed32076be4635649/src/protocols/kerberos.rs#L201
However, I believe they need to be thoroughly verified.

src/graphql/client/schema/schema.graphql
```diff
@@ -1927,24 +1927,60 @@ type SecuLogRawEventEdge {
...
+  # File size
   fileSize: StringNumberU64!
+
+  # Resource type.
   resourceType: Int!
+
+  # File ID.
   fid: Int!
+
+  # Create time.
```
danbi2990, line 1974: FWIW, it seems that the unit is 100 nanoseconds.
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/2c57429b-fdd4-488f-b5fc-9e4cf020fcdf

src/graphql/client/schema/schema.graphql
sophie-cluml: I agree. But we should not share private links in public spaces.

sophie-cluml (approved):

Issue #1015, opened 2025-02-07
Correctly modify Giganto's config handling behavior


sophie-cluml:
The following changes need to be made:  

- Giganto should always operate based on a local configuration file.  
  - The reason for this is that, unlike other agents, Giganto functions as a server module. From a configuration perspective, this means that Giganto’s config is not stored at a remote location. Giganto stores its own configuration.
  - The necessary modifications include:  
    - When launching Giganto, specifying the configuration file location should be mandatory.
    - When a `setConfig` request is received, the configuration file at the specified location should be directly overwritten. Detailed error handling for potential issues during this process should be addressed in a separate issue.  
    - Rename the `setConfig` API to `updateConfig`. Currently, when this API is called, it not only sets the configuration but also applies it to Giganto (i.e., set & reload). Renaming it to `updateConfig` better conveys this behavior while keeping it simple.  
    - Remove any configuration-related code that does not align with this new approach. For example, since Giganto will now always rely on a local configuration file, idle mode can no longer exist. Any code assuming idle mode should be removed accordingly.  

These modifications will ensure a more consistent and reliable configuration handling approach for Giganto.

This issue is created from a discussion with @sehkone.

PR #1023, opened 2025-02-14
Modify configuration management

sophie-cluml:
Closes: #1015

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,39 +24,45 @@ processing and real-time analytics.
 You can run Giganto by invoking the following command:
 
 ```sh
-giganto --cert <CERT_PATH> --key <KEY_PATH> --ca-certs <CA_CERT_PATH> \
---ca-certs <CA_CERT_PATH>
-```
-
-If you want to run Giganto with local configuration file,
-
-```sh
 giganto -c <CONFIG_PATH> --cert <CERT_PATH> --key <KEY_PATH> --ca-certs \
-<CA_CERT_PATH> --ca-certs <CA_CERT_PATH>
+<CA_CERT_PATH>[,<CA_CERT_PATH>...] [--log-dir <LOG_DIR>]
 ```
 
 ### Arguments
 
-- `<CONFIG_PATH>`: Path to the TOML configuration file (optional when running in
-  remote mode).
-- `<CERT_PATH>`: Path to the certificate file (required).
-- `<KEY_PATH>`: Path to the private key file (required).
-- `<CA_CERTS_PATH>`: Path to the CA certificates file (required).
+<!-- markdownlint-disable -->
+
+| Name             | Description                                               | Required |
+| ---------------- | --------------------------------------------------------- | -------- |
+| `<CONFIG_PATH>`  | Path to the TOML configuration file.                      | Yes      |
+| `<CERT_PATH>`    | Path to the certificate file.                             | Yes      |
+| `<KEY_PATH>`     | Path to the private key file.                             | Yes      |
+| `<CA_CERT_PATH>` | Path to the CA certificates file.                         | Yes      |
+| `<LOG_DIR>`      | Path to the directory where the log files will be stored. | No       |
+
+<!-- markdownlint-enable -->
+
+#### Notes on Arguments
+
+- The `--ca-certs` argument accepts multiple values, separated by commas. You
+  can also repeat the argument to specify multiple CA certificates.
+- Logging behavior based on the `--log-dir` argument is as follows:
+  - If `<LOG_DIR>` is not provided, logs are written to stdout using the tracing
+    library.
+  - If `<LOG_DIR>` is provided and writable, logs are written to the specified
+    directory using the tracing library.
+  - If `<LOG_DIR>` is provided but not writable, Giganto will terminate.
+  - Any logs generated before the tracing functionality is initialized will be
+    written directly to stdout or stderr using `println`, `eprintln`, or
+    similar.
 
 ### Example
-
-- Run Giganto with remote server configuration.
-
-```sh
-giganto --cert /path/to/cert.pem --key /path/to/key.pem \
---ca-certs /path/to/ca_cert.pem
-```
 
 - Run Giganto with local configuration file and multiple CA certificates.
 
 ```sh
 giganto -c path/to/config.toml --cert /path/to/cert.pem --key /path/to/key.pem \
---ca-certs /path/to/ca_cert1.pem --ca-certs /path/to/ca_cert2.pem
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -23,25 +23,36 @@ processing and real-time analytics.
...
-giganto --cert <CERT_PATH> --key <KEY_PATH> --ca-certs <CA_CERT_PATH> \
---ca-certs <CA_CERT_PATH>
-```
-
-If you want to run Giganto with local configuration file,
-
 ```sh
 giganto -c <CONFIG_PATH> --cert <CERT_PATH> --key <KEY_PATH> --ca-certs \
-<CA_CERT_PATH> --ca-certs <CA_CERT_PATH>
+<CA_CERT_PATH> --ca-certs <CA_CERT_PATH> --log-dir <LOG_DIR>
```
sehkone, line 28: Since `<LOG_DIR>` can be not given, using `[ ]` might help, which means optional.

sehkone: Does `--ca-certs` allow multiple paths?

src/settings.rs
```diff
@@ -38,55 +45,56 @@ pub struct Args {
...
+}
 /// The application settings.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct Config {
+    #[serde(default, deserialize_with = "deserialize_peer_addr")]
+    pub addr_to_peers: Option<SocketAddr>, // IP address & port for peer connection
+    pub peers: Option<HashSet<PeerIdentity>>,
+
+    #[serde(flatten)]
+    pub updatable: UpdatableConfig,
```
sehkone, line 72: Because `addr_to_peers` and `peers` are not always immutable, I don't think the name `UpdatableConfig` is not ideal. This can mislead people into thinking both are immutable.

sehkone: Does `ConfigVisible` sound good to you?

src/graphql/status.rs
```diff
@@ -187,44 +162,46 @@ impl StatusQuery {
...
     #[allow(clippy::unused_async)]
-    async fn set_config(&self, ctx: &Context<'_>, draft: String) -> Result<bool> {
-        let is_local = ctx.data::<bool>()?;
-        let TracingEnabled(tracing_enabled) = ctx.data::<TracingEnabled>()?;
-        let tracing_enabled = *tracing_enabled;
-
-        if *is_local {
-            warn!("The request to the GraphQL API to change the configuration was ignored because this is running with a local configuration.");
-            return Err("Config is local".to_string().into());
+    async fn update_config(&self, ctx: &Context<'_>, draft: String) -> Result<bool> {
```
sehkone, line 176: Since the `draft` is stored in the manager, I don't Giganto needs this term. It should be `new_config`. In addition, this API should receive the old `config` to prevent conflicts from concurrent requests.

sophie-cluml: Done

src/settings.rs
sophie-cluml: Done

README.md
sophie-cluml: > Since <LOG_DIR> can be not given, using [ ] might help, which means optional:

Done


> Does --ca-certs allow multiple paths?

Yes, I added the following line to the README.

```
- The `--ca-certs` argument can be repeated to specify multiple CA certificates.
```

sehkone: I think there should be a conventional way to represent multiple values.

README.md
```diff
@@ -111,14 +121,11 @@ be appropriate.
 
 These values assume you've used all the way up to level 6, so the actual values may
 change if you want to grow your data further at the level base.
-So if it's less than `512`MB, it's recommended to set default value of `512`MB.
+So if it's less than 512MB, it's recommended to set default value of 512MB.
```
sehkone, line 124: To write `512 MB` is more standard.

src/graphql/status.rs
```diff
@@ -187,44 +162,61 @@ impl StatusQuery {
...
     #[allow(clippy::unused_async)]
-    async fn set_config(&self, ctx: &Context<'_>, draft: String) -> Result<bool> {
-        let is_local = ctx.data::<bool>()?;
-        let TracingEnabled(tracing_enabled) = ctx.data::<TracingEnabled>()?;
-        let tracing_enabled = *tracing_enabled;
-
-        if *is_local {
-            warn!("The request to the GraphQL API to change the configuration was ignored because this is running with a local configuration.");
-            return Err("Config is local".to_string().into());
+    async fn update_config(&self, ctx: &Context<'_>, old: String, new: String) -> Result<bool> {
```
sehkone, line 176: When does this return `Ok(false)`?

README.md
sophie-cluml: Do you mean comma separated values?

src/graphql/status.rs
sophie-cluml: Currently it does not return `Ok(false)`.

README.md
sophie-cluml: Thank you, done.

src/graphql/status.rs
sehkone: Given that, I think returning `Result<String>` would be better.

sophie-cluml: Done

README.md
sophie-cluml: I changed it to accept multiple values with commas. I'd appreciate if you could let me know if this is what you have in mind.

src/graphql/status.rs
```diff
@@ -187,46 +162,63 @@ impl StatusQuery {
...
+            tx_clone.send(new_config).await.map_err(|e| {
+                error!("Failed to send config: {e:?}");
                 "Failed to send config".to_string()
             })
         });
-        log(tracing_enabled, Level::INFO, "Draft applied.");
+        info!("New config is applied.");
 
-        Ok(true)
+        Ok("Success")
```
sehkone, line 221: I think returning the new `config` would be more consistent with our codebase. If this method were to return `true` or `false` depending on the context, `Result<bool>` would be perfect. However, since it never returns `false`, we need to use `String` instead. Just like APIs such as `update_account`, which return an `ID`, I believe meaningful data should be returned. In this context, the only meaningful data is the new `config`.

sophie-cluml: Done

README.md
sophie-cluml: Done

sehkone (approved):

README.md
sehkone: @sophie-cluml,

I apologize for the incorrect alternative I provided. I have revised it to `--ca-certs <CA_CERT_PATH>[,<CA_CERT_PATH>,...]` for better clarity. The last comma is necessary.

PR #1043, opened 2025-02-28
Automate GraphQL schema generation

### LINKED ISSUE #1042: Automate GraphQL schema generation
Currently, whenever a GraphQL query changes, the corresponding GraphQL schema file must also be updated in the specified directory.
The steps to update the schema file are as follows.

1. Run the GraphQL playground.
2. Download the schema file.
3. Place it in the `src/graphql/client/schema` directory.

This issue aims to reduce the inconveniences and minimize the potential for human error.

Related discussion: https://github.com/aicers/giganto/discussions/992#discussioncomment-12245142

danbi2990:
Added a test to ensure that GitHub Actions prevents the merging of pull requests with an outdated schema.

Closes #1042

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,18 @@ cargo run -- -c tests/config.toml --cert tests/certs/node1/cert.pem \
 --key tests/certs/node1/key.pem --ca-certs tests/certs/ca_cert.pem
 ```
 
+## Contributing
+
+To run Giganto, the `src/graphql/client/schema/schema.graphql` file is required.
+This file is used to enable GraphQL client functionality, which is necessary for
+clustering. The file is not manually maintained, but generated from the relevant
+GraphQL source code. Therefore, if the source code affecting the GraphQL API has
+been modified, you need to update the schema by running:
+
+```sh
+cargo run --features=export-graphql-schema
+```
+
 ## License
 
 Copyright 2022-2025 ClumL Inc.
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "export-graphql-schema", allow(unused))]
+
 mod graphql;
 mod ingest;
 mod peer;
@@ -52,11 +54,16 @@ use crate::{
 const ONE_DAY: Duration = Duration::from_secs(60 * 60 * 24);
 const WAIT_SHUTDOWN: u64 = 15;
 
+#[allow(dead_code)]
+#[cfg(any(debug_assertions, feature = "export-graphql-schema"))]
+const GRAPHQL_SCHEMA_PATH: &str = "src/graphql/client/schema/schema.graphql";
+
 pub type PcapSensors = Arc<RwLock<HashMap<String, Vec<Connection>>>>;
 pub type IngestSensors = Arc<RwLock<HashSet<String>>>;
 pub type RunTimeIngestSensors = Arc<RwLock<HashMap<String, DateTime<Utc>>>>;
 pub type StreamDirectChannels = Arc<RwLock<HashMap<String, UnboundedSender<Vec<u8>>>>>;
 
+#[cfg(not(feature = "export-graphql-schema"))]
 #[allow(clippy::too_many_lines)]
 #[tokio::main]
 async fn main() -> Result<()> {
@@ -467,4 +474,46 @@ async fn wait_for_task_shutdown(
         let _ = tokio::join!(ingest_task_handle, publish_task_handle);
     }
     let _ = retain_task_handle.join();
+}
+
+#[cfg(feature = "export-graphql-schema")]
+fn main() -> Result<()> {
+    use async_graphql::{EmptySubscription, Schema};
+
+    use crate::{
+        graphql::{Mutation, Query},
+        GRAPHQL_SCHEMA_PATH,
+    };
+
+    fs::write(
+        GRAPHQL_SCHEMA_PATH,
+        Schema::build(Query::default(), Mutation::default(), EmptySubscription)
+            .finish()
+            .sdl(),
+    )
+    .context("Failed to write the GraphQL schema")?;
+    println!("Successfully exported GraphQL schema.");
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use async_graphql::{EmptySubscription, Schema};
+
+    use crate::{
+        graphql::{Mutation, Query},
+        GRAPHQL_SCHEMA_PATH,
+    };
+
+    #[test]
+    fn graphql_schema_should_be_up_to_date() {
+        let expected = Schema::build(Query::default(), Mutation::default(), EmptySubscription)
+            .finish()
+            .sdl();
+        let actual = std::fs::read_to_string(GRAPHQL_SCHEMA_PATH).unwrap();
+        assert!(
+            expected == actual,
+            "GraphQL schema is outdated. Please run `cargo run --features=export-graphql-schema`"
+        );
+    }
 }
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -27,6 +27,7 @@ const DEFAULT_MAX_MB_OF_LEVEL_BASE: u64 = 512;
 const DEFAULT_NUM_OF_THREAD: i32 = 8;
 const DEFAULT_MAX_SUB_COMPACTIONS: u32 = 2;
 
+#[cfg(not(feature = "export-graphql-schema"))]
 #[derive(Parser, Debug)]
 #[command(version)]
 pub struct Args {
@@ -59,6 +60,9 @@ pub struct Args {
     #[arg(long)]
     pub repair: bool,
 }
+
+#[cfg(feature = "export-graphql-schema")]
+pub struct Args {}
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Settings {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,4 @@ url = "2"
 [features]
 default = ["benchmark"]
 benchmark = []
+export-graphql-schema = []
--- a/src/graphql/client/schema/schema.graphql
+++ b/src/graphql/client/schema/schema.graphql
@@ -1,2464 +1,2855 @@
-# Indicates that an Input Object is a OneOf Input Object (and thus requires
-#                         exactly one of its field be provided)
-directive @oneOf on INPUT_OBJECT
 
-# Represents an event extracted from the BOOTP protocol.
+"""
+Represents an event extracted from the BOOTP protocol.
+"""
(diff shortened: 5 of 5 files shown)


sophie-cluml:
@danbi2990 What do you think about having a quick sync-up for this today?

danbi2990:
> @danbi2990 What do you think about having a quick sync-up for this today?

Sure.

src/settings.rs
```diff
@@ -57,6 +57,10 @@ pub struct Args {
     /// Enable the repair mode.
     #[arg(long)]
     pub repair: bool,
+
+    /// Export GraphQL schema.
+    #[arg(long)]
+    pub export_graphql_schema: bool,
```
sophie-cluml, line 63: I don't think we can expose this argument to our customers. As we discussed in a different communication channel, let's disable this argument in some cases (such as for production), and enable  this argument in some other cases (such as during development). And if possible, I think we can guide users to exclusively use `export_graphql_schema` argument, without specifying other arguments.

danbi2990: Testing confirms that `--export-graphql-schema` option...
- [x] Should not appear in `--help` command
  - Hidden in release builds, visible in debug builds
- [x] Should not be runable in release builds
- [x] Should be callable on its own without other arguments
- [x] Should preserve the existing CLI and --help text

sophie-cluml: I know I originally suggested using `cfg(debug_assertions)`, and it somewhat made sense back then. But considering our recent efforts to align debug and release behavior the same, I’m starting to think a dedicated Cargo feature might offer clearer control. I would love to hear your thoughts. What do you think about this?

sophie-cluml: @danbi2990 Could you check above comment please?

danbi2990: I agree with the idea and it should be simple change to add a new feature for it. Let me fix it.

By the way, I had a sense that we were moving away from using build type to control application logic, but I wasn't aware that the goal was to "align debug and release behavior." Could you elaborate on this for better understanding or share a relevant discussion link, if there is one?

sophie-cluml: Since the link is private, let me share it in another channel.

danbi2990: Applied the suggested adjustment.
Considering the Open-Closed principle, a new `Args` struct and main function were added rather than modifying the existing ones.

sophie-cluml (commented):
I think the change looks good, and I think a description in README might be very helpful for users and contributors. What do you think about adding a description?

danbi2990:
> I think the change looks good, and I think a description in README might be very helpful for users and contributors. What do you think about adding a description?

Added Contributing section to README.

sophie-cluml (approved):

sophie-cluml:
This looks good to me. Since @sehkone expressed interest in this topic during the previous discussion  https://github.com/aicers/giganto/discussions/992#discussioncomment-12048985, I would like to invite him to review this PR in case he's interested in the details. If there's no response, I plan to proceed with merging this PR a few days after this comment.

sehkone:
> This looks good to me. Since @sehkone expressed interest in this topic during the previous discussion [#992 (comment)](https://github.com/aicers/giganto/discussions/992#discussioncomment-12048985), I would like to invite him to review this PR in case he's interested in the details. If there's no response, I plan to proceed with merging this PR a few days after this comment.

I’ll take a look and follow up shortly.

src/main.rs
```diff
@@ -52,11 +54,16 @@ use crate::{
 const ONE_DAY: Duration = Duration::from_secs(60 * 60 * 24);
 const WAIT_SHUTDOWN: u64 = 15;
 
+#[allow(dead_code)]
```
sehkone, line 57: Is this line required?

README.md
```diff
@@ -130,6 +130,18 @@ cargo run -- -c tests/config.toml --cert tests/certs/node1/cert.pem \
 --key tests/certs/node1/key.pem --ca-certs tests/certs/ca_cert.pem
 ```
 
+## Contributing
```
sehkone, line 133: From what I understand, this GraphQL generation is essential for clustering Giganto instances. If that’s the case, it should be treated as a mandatory step in every PR that modifies the schema. However, the current “Contributing” section doesn’t convey this level of importance—nor is the terminology used there appropriate.

sehkone:
@sophie-cluml and @danbi2990 

Based on a bit of research I did over the last few minutes, I think this new approach might be viable—and potentially better than the one taken in this PR, if it works as expected:

- Add `build.rs`.
- Have it automatically detect whether any API-related code has changed.
- If so, trigger a separate binary—say `gen_schema`—that shares the API module of Giganto.
- This binary generates the schema using the `sdl` method.

If this is feasible to implement, I believe it would provide a more automated and seamless process, reducing the potential for human error.

Have you already considered and discarded this approach for any reason? If not, I’d like to ask you to take a look and let me know your thoughts.

danbi2990:
> Have you already considered and ruled out this approach for any reason? If not, I’d really appreciate it if you could take a look and share your thoughts.

Sophie and I previously discussed this approach, and I've revisited it through further research. While it is technically feasible, there are several obstacles that make it less straight forward.

# 1. Deadlock from concurrent `cargo run` executions

- When a user runs `cargo run`, the `build.rs` script also triggers `cargo run --bin gen_schema`.
- This results in two concurrent `cargo run` processes, which can lead to a deadlock due to Cargo's internal file locking.

To avoid this, we could precompile the `gen_schema` binary and execute the compiled artifact directly in `build.rs`. However, this may require a pre-build script to ensure `gen_schema` is built before the main build starts.

# 2. Crate restructuring

To generate the GraphQL schema file, `graphql` module must be shared between two binaries. This can be acheived by moving shared logic into `lib.rs`. However, since the current `lib.rs` is used for proc_macros, it must be relocated to a dedicated crate first.

A simplified structured layout would look like:

src/

- main.rs
- gen_schema.rs
- lib.rs

giganto_proc_macro/
- src/lib.rs
- Cargo.toml

This restructuring would require several code changes:
- Update all `crate::...` paths to `giganto::...`, reflecting the move of the module tree into lib.rs.
- Move all `mod ...` declarations from `main.rs` into `lib.rs`.
- Move all public functions in `main.rs` into `lib.rs`.
- Relocate all procedural macros currently in `lib.rs` to `giganto_proc_macro/src/lib.rs`.

While this approach is clearly more verbose than the current PR, it could be a worthwhile opportunity to modularize and clean up the codebase. I’d appreciate your thoughts on whether this tradeoff is acceptable or if we should consider an alternative path.

sehkone:
> # 1. Deadlock from concurrent `cargo run` executions
> * When a user runs `cargo run`, the `build.rs` script also triggers `cargo run --bin gen_schema`.
> * This results in two concurrent `cargo run` processes, which can lead to a deadlock due to Cargo's internal file locking.

- Is it true that `cargo run` triggers not only `main.rs` but also source files for other binaries?
- Even if that’s the case, couldn’t we avoid this by using `cargo build` instead of `cargo run`?  Or perhaps by using `cargo run` with specific options?

danbi2990:
> * Is it true that `cargo run` triggers not only `main.rs` but also source files for other binaries?

What I meant is that the `build.rs` file triggers the other binary, `gen_schema`. While the `graphql` module is compiled into the both binaries, the source files themselves are not related to the deadlock. 

The deadlock occurs because the first `cargo`   process locks the target directory, and the second `cargo` process attempts to acquire a lock on the same target directory.

> * Even if that’s the case, couldn’t we avoid this by using `cargo build` instead of `cargo run`?  Or perhaps by using `cargo run` with specific options?

Since both `cargo build` and `cargo run` lock the target directory, the issue is not specific to the command used. 

However, as you said, there's `--target-dir` option in the `cargo` command allows each `cargo` process to use a separate target directory, which can prevent the deadlock when combined with an additional environment variable as follows:

```rust
// build.rs example
fn main() {
    if env::var("BUILD_GEN_SCHEMA").is_err() {  // env var is required to avoid recursion
        let status = Command::new("cargo")
            .env("BUILD_GEN_SCHEMA", "1")
            .args(["run", "--bin", "gen_schema", "--target-dir", "target_gen_schema"])
            .status()
            .expect("Failed to run gen_schema");
        assert!(status.success(), "gen_schema failed to run");
    }

    println!("cargo::rerun-if-changed=src/graphql.rs");
}
```

sehkone:
The key point I’d like to emphasize is that Giganto developers should be able to compile and run the application without any additional steps for generating the schema.

To meet this requirement, it seems we inevitably need to take advantage of `build.rs`.  Whether we use `cargo run` or `cargo build` within `build.rs` is optional—alternatively, we can simply run the already-compiled `gen_schema` binary.

As for creating a separate repository for the macro-related code, I believe you’re better informed on that than I am.  I’d appreciate it if you could explore and suggest an appropriate approach.

danbi2990:
> As for creating a separate repository for the macro-related code, I believe you’re better informed on that than I am. I’d appreciate it if you could explore and suggest an appropriate approach.

I'd like to propose restructuring the current codebase into two crates: `giganto-proc-macro` and `giganto-core`.
- Move `lib.rs` to the `giganto-proc-macro` crate.
- Move all source files except `main.rs` to the `giganto-core` crate.

The root `Cargo.toml` should be updated to use a `[workspace]` configuration. In this structure, a `gen_schema` function will be added to the `giganto-core::graphql` module. Then, the root-level `build.rs` will invoke this function. To enable this, `giganto-core` should be added under [build-dependencies] in the root `Cargo.toml`.

The idea is inspired by the [`tonic-build`](https://github.com/hyperium/tonic/blob/master/tonic-build/README.md) crate, which generates service stubs and proto definitions from `.proto` files. I believe our use case is quite similar.

A limitation is that changes to any part of the `giganto-core` crate will trigger `build.rs`, not just changes to the `graphql` module.

# Unchosen Alternatives

##  [`xtask`](https://github.com/matklad/cargo-xtask/)

This is a Rust-based alternative to tools like `make`, `npm run`, or shell scripts, commonly used in other Rust projects. However, I decided against it for the following reasons:
- `xtasks` tasks are not integrate into Cargo lifecycle, so schema generation won't happen automatically.
- We currently have only one task, which is sufficiently handled by `build.rs`.

## [Generating code via a test](https://old.reddit.com/r/rust/comments/tp8tmn/blog_post_self_modifying_code/i2ad6a0/)

This approach is similar to our current PR, except generation is done via a test. It's used in [rust-analyer](https://github.com/rust-lang/rust-analyzer/blob/96c11f5fd6bb66d60932275ff052025ea6073fb6/crates/syntax/src/tests/sourcegen_ast.rs#L21). While it's simple, it's not as straightforward.

danbi2990:
This discussion continues in https://github.com/aicers/giganto/discussions/1099

PR #1077, opened 2025-04-24
Change logging behavior to depend on log_path

### LINKED ISSUE #1075: Log behavior depends on `log_path`, not build config
If `log_path` is set, logs should be written to that path; otherwise, they should be printed to stdout.

Goder-0:
## Related Issue
- close #1075

## PR Description
- Previously, logging behavior depended on both `log_path` and the build type(debug/release).
- Removed the dependency on the build type.
- Logging is now controlled solely by `log_path`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Logging behavior no longer considers the debug mode. Previously, logging
+  behavior depended on both `log_path` and the build type(debug/release). Now,
+  whether logs are written to stdout or to a file is determined solely by the
+  presence of `log_path`.
 - Giganto now creates a backup before overwriting the config via the
   `updateConfig` GraphQL API. If reading the config fails, it automatically
   restores from the backup, both during `updateConfig` calls and at startup.
--- a/src/main.rs
+++ b/src/main.rs
@@ -99,7 +99,7 @@ async fn main() -> Result<()> {
 
     let cfg_path = settings.cfg_path.clone();
 
-    let _guards = init_tracing(args.log_path.as_deref())?;
+    let _guard = init_tracing(args.log_path.as_deref())?;
 
     if args.repair {
         repair_db(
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Logging behavior no longer considers the debug mode. Previously, logging
+  behavior depended on both `log_path` and the build type(debug/release). Now,
+  whether logs are written to stdout or to a file is determined solely by the
+  value of `log_path`.
```
danbi2990, line 15: How about changing "value" to "presence"?
In my opinion, it's determined by the presence of `log_path`, not its value.

Goder-0: Thank you for the suggestion. Done!

src/main.rs
```diff
@@ -417,37 +417,29 @@ fn new_peers_data(peers_list: Option<HashSet<PeerIdentity>>) -> (Peers, PeerIden
 ///
 /// Returns an error if the log file cannot be opened in the `log_path` path in the
 /// local configuration.
-fn init_tracing(log_path: Option<&Path>) -> Result<Vec<WorkerGuard>> {
-    let mut guards = vec![];
-
-    let file_layer = if let Some(log_path) = log_path {
-        let file_path = log_path.to_path_buf();
+fn init_tracing(log_path: Option<&Path>) -> Result<WorkerGuard> {
```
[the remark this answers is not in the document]
Goder-0: I didn’t realize there was a comment. I’ve applied the change now!

danbi2990 (approved):

Goder-0:
I think it would be best to perform the rebase after #1078 has been resolved.

sophie-cluml (approved):

sehkone:
@Goder-0,

In the future, I think it would be better to place information like “Closes …” at the bottom of the body.

Goder-0:
> @Goder-0,
> 
> In the future, I think it would be better to place information like “Closes …” at the bottom of the commit body.

@sehkone Thank you for the suggestion.
After reviewing the related documentation on conventions, I found that Korean documents place such information earlier, but in English documents (as referenced in the shared PR), it is typically placed at the bottom of the commit body.
I will align with this convention going forward.

Issue #1098, opened 2025-05-27
Consistent Handling of Search Range Boundary Value


sophie-cluml:
### Background 

https://github.com/aicers/giganto/discussions/1093

### Task

- Revert #593
- Ensure consistent handling of search range boundary values as outlined in #136 
  - Review the codebase for any other cases where this consistency is not maintained, and update them accordingly.

PR #1113, opened 2025-06-10
Ensure consistent search range boundary handling

kimhanbeom:
- Modify `check_address` and `check_port` to ensure consistent search range boundaries for time, address, and port.
  (There are no additional modifications needed beyond the current changes in the codebase.)

Close: #1098

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -769,37 +769,35 @@ fn write_run_tcpdump(packets: &Vec<pk>) -> Result<String, anyhow::Error> {
 }
 
 fn check_address(filter_addr: Option<&IpRange>, target_addr: Option<IpAddr>) -> Result<bool> {
-    if let Some(ip_range) = filter_addr {
-        if let Some(addr) = target_addr {
-            if let Some(start) = ip_range.start.as_deref() {
-                if let Some(end) = ip_range.end.as_deref() {
-                    if addr >= start.parse::<IpAddr>()? && addr < end.parse::<IpAddr>()? {
-                        return Ok(true);
-                    }
-                    return Ok(false);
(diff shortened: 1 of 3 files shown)


src/graphql.rs
```diff
@@ -791,11 +794,20 @@ fn check_address(filter_addr: Option<&IpRange>, target_addr: Option<IpAddr>) ->
...
+            };
+
+            let end = if let Some(end) = port_range.end {
+                port >= end
+            } else {
+                false
+            };
+
+            if start || end {
+                return false;
```
sophie-cluml, line 810: I think it might be  expressed concisely using match:

```
match (filter_port, target_port) {
    (Some(port_range), Some(port)) => ...,
    _ => true,
}
```

Also, in the current code, the `start` and `end` variables are used in a way that represents "does not pass start check" and "does not pass end check," which makes the naming somewhat harder to follow. The logic itself seems correct, but could you please consider adjusting the variable names and structure to use positive logic instead? I believe this would make the code easier to read and maintain.

kimhanbeom: done

src/graphql.rs
```diff
@@ -769,37 +769,45 @@ fn write_run_tcpdump(packets: &Vec<pk>) -> Result<String, anyhow::Error> {
...
+                true
+            };
+
+            let ends_before = if let Some(end) = port_range.end {
+                port < end
+            } else {
+                true
+            };
+
+            starts_after_or_at && ends_before
```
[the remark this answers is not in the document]
kimhanbeom: done

sophie-cluml (approved):

PR #1133, opened 2025-07-08
Add cluster feature and modularize common logic

### LINKED ISSUE #1110: Modularize GraphQL Client Logic under `graphql::client`
Refactor the codebase to consolidate client-related logic under the `graphql::client` module wherever possible. For components that cannot be moved (e.g., due to shared utilities or structs), implement conditional compilation using `cfg`, or `cfg_attr` or other methods, and a new Cargo feature (personally proposed name: "cluster").

This separation will enable two distinct build modes:

- Server-only build (with `--no-default-features`): Excludes client-specific code. Produces a Giganto binary that operates only as a standalone instance, fully functional as a single Giganto, but unable to make GraphQL API calls to other Gigantos.
- Full build (default build, with personally proposed "cluster" cargo feature): Includes client logic for cluster functionality. Produces a Giganto binary capable of making GraphQL API calls to other Gigantos.

After refactoring, the codebase must support both build modes without breaking existing functionality.

danbi2990:
Testing confirms that

- DNS events are detected by the semi-supervised engine in both cluster and standalone modes.
- DNS events are listed correctly in the UI in both cluster and standalone modes.
- Communication between Giganto and time-series-generator works as expected.

Closes #1110


kimhanbeom:
It might be good to add a Clippy check to ci.yml based on the cluster feature.
```yml
      - name: Clippy (full)
        run: cargo clippy --tests -- -D warnings
      - name: Clippy (Server-only)
        run: cargo clippy --no-default-features --features benchmark --tests -- -D warnings
```

kimhanbeom (approved):

sophie-cluml (commented):
Thank you for your work on this PR. Each cluster and non-cluster build appear to be working as expected, but I have a few questions and suggestions:

- Regarding the necessity of the comm module : Could you elaborate on the decision to introduce a new `comm` module? Was this done in consideration of future issues that will follow after #1110? I would like to keep the project structure simpler as much as I can.
- Naming of comm module: I assume it stands for common, referring to shared functionality between standalone and cluster modes). That said, since the module currently includes modules like ingest, publish, and peer, I wonder if there are other shared functionalities that haven't been moved into the comm module. If that's the case, it might be worth reconsidering the module name. Alternatively, this may circle back to my earlier question about whether a separate comm module is necessary, or if the project structure could be kept simpler without it.
- Use of mod.rs :  It seems that the comm module uses the older mod.rs pattern. To align with current project conventions, could you consider flat file style (e.g., comm.rs)?
- Scope of #[allow(unused_variables)]: I understand the necessity of this allow attribute, but rather than applying #[allow(unused_variables)] at a function level, I suggest limiting it to specific variables that are actually unused, for improved code clarity.

Please let me know if there are intentional choices that I might have missed. Thank you again for the thorough work.

danbi2990:
Updated the module structure and attribute.

sophie-cluml (approved):

sophie-cluml:
Thank you for the work.

PR #1151, opened 2025-07-24
Automate GraphQL schema generation

### LINKED ISSUE #1112: Automate GraphQL Schema Generation in Build Pipeline
Implement Schema Generation Step in Build Pipeline

### Issue Dependency

Depends on #1110 and #1111.

### Details

Integrate the GraphQL schema generation step into the Giganto build pipeline using build.rs. The workflow for example is as follows:

1. User Action: Runs `cargo build` (full build).
2. Build Script (build.rs): Compiles the schema generation binary.
3. Build Script (build.rs):  Executes the binary to generate schema.graphql at a predefined path.

The process should be fully automated. In other words, running `cargo build` should be sufficient for both schema generation and the final build.

danbi2990:
Closes #1112

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,29 @@
+use std::{env, fs, process::Command};
+
+fn main() {
+    if env::var("BUILD_GEN_SCHEMA").is_err() && env::var("PROFILE").unwrap_or_default() == "dev" {
+        let status = Command::new("cargo")
+            .env("BUILD_GEN_SCHEMA", "1")
+            .args([
+                "run",
+                "--bin",
+                "gen_schema",
+                "--no-default-features",
+                "--target-dir",
+                "target_gen_schema",
+            ])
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -0,0 +1,29 @@
...
+                "--no-default-features",
+                "--target-dir",
+                "target_gen_schema",
+            ])
+            .status()
+            .expect("Failed to execute `cargo run` for schema generation");
+        assert!(status.success(), "Failed to generate GraphQL schema");
+    }
+
+    // Re-run only if non-client files in src/graphql change
```
sophie-cluml, line 20: I think src/graphql.rs should be included for the rerun-if-changed.

danbi2990: Done.

sophie-cluml (approved):

PR #1168, opened 2025-08-08
Fix clippy error for Rust 1.89.0

kimhanbeom:
Close: #1167

### DIFF
--- a/src/graphql/client/derives.rs
+++ b/src/graphql/client/derives.rs
@@ -79,12 +79,15 @@ pub struct BootpRawEvents;
 )]
 pub struct DhcpRawEvents;
 
+// Currently, `PeriodicTimeSeries` is not used in cluster mode. However, this API may be extended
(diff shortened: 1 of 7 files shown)


src/graphql/client/derives.rs
```diff
@@ -79,12 +79,16 @@ pub struct BootpRawEvents;
 )]
 pub struct DhcpRawEvents;
 
+/// Currently, `PeriodicTimeSeries` is not used in cluster mode.
+/// However, this GraphQL API may be extended for use in the future.
+/// For now, `#[allow(dead_code)]` is applied to suppress clippy warnings.
```
sophie-cluml, line 84: For this specific case, I think it might be clearer to start the comment with two slashes (//), because the content appears to be quite temporary remark intended for developers' internal understanding.

kimhanbeom: done

sophie-cluml (approved):

Issue #1325, opened 2025-11-05
Event transmission fails when Giganto DB is empty


pott-cho:
### Description

When the Giganto DB does not contain the sensor information referenced by the events that the Data Broker is trying to send,
the Data Broker fails to transmit those events.
In other words, if the Giganto database is completely empty, no events can be successfully inserted into it via the Data Broker.

### Steps to Reproduce

1. Initialize a clean Giganto DB and start it using the latest branch.
2. Start the Data Broker (latest branch) and send `event1_log`.csv.
3. Check via GraphQL whether the event was received by Giganto → no data observed.

kimhanbeom:
## Cause of the Issue
- In issue #1304, the `check_sensors_conn` logic was modified to manage connections only for sensor applications.
- As a result, when data-broker sends data, its hostname(extracted from the certificate) is not stored in `IngestSensors`.
- When querying events via GraphQL, the `is_current_giganto_in_charge` function checks whether the current Giganto instance should handle the event. Since the data-broker's hostname is missing from `IngestSensors`, the event is skipped, and the query returns empty even though the data is actually stored.
    ```rust
    pub(crate) async fn is_current_giganto_in_charge(ctx: &Context<'_>, sensor_filter: &str) -> bool {
        let ingest_sensors = ctx.data_opt::<IngestSensors>();
        match ingest_sensors {
            Some(ingest_sensors) => ingest_sensors.read().await.contains(sensor_filter),
            None => false,
        }
    }
    ```
- Similar issue is expected to occur in the publish module as well when receiving requests for data from data-broker.
   ```rust 
   async fn is_current_giganto_in_charge(ingest_sensors: IngestSensors, sensor: &str) -> bool {
       ingest_sensors.read().await.contains(sensor)
   }
   ```

## Proposed Solution
- The original goal was to support [all connected sources via ingest](https://github.com/aicers/giganto/issues/40), so the changes made in issue #1304 were unnecessary.
- Therefore, we should revert the logic to its previous state, excluding the `PcapSensor` part (which is specific to pcap extraction and should remain unchanged).

PR #1326, opened 2025-11-05
Revert logic to handle all ingest sources

kimhanbeom:
Fix: #1325

### DIFF
--- a/src/comm/ingest.rs
+++ b/src/comm/ingest.rs
@@ -60,7 +60,7 @@ const NO_TIMESTAMP: i64 = 0;
 const SENSOR_INTERVAL: u64 = 60 * 60 * 24;
 const INGEST_VERSION_REQ: &str = ">=0.26.0-alpha.6,<0.26.0-alpha.8";
 
-type SensorInfo = (String, DateTime<Utc>, ConnState);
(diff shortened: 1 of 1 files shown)


src/comm/ingest.rs
```diff
@@ -1126,7 +1131,7 @@ async fn check_sensors_conn(
                 }
             }
 
-            Some((sensor_key, time_val, conn_state)) = rx.recv() => {
+            Some((sensor_key, time_val, conn_state, sensor)) = rx.recv() => {
```
sophie-cluml: I think the term sensor in the `sensor_key` is used to mean a sensor encompassing any types of sensors including 'the' sensor and our `reproduce` application, covering the full scope implied by "sensor." However, the last element named `sensor` seems to refer specifically to a sensor that collects pcap data. I think using the same term for different meanings can cause confusion, so I would like to suggest that we distinguish them, for example by renaming the `sensor` to something like `pcap_sensor` or `is_pcap_sensor` to clarify its intended meaning.

In the same perspective, the variable `should_register_sensor` used around L204 seems to intend that "this sensor should be registered as a pcap sensor." I think it would be good to reconsider the variable naming to something like `is_pcap_sensor` or any other better fitting name that clearly conveys this meaning.

Issue #1335, opened 2025-11-15
Update giganto-client to `6fe7654`, and resolve related migration issues


sophie-cluml:
As of https://github.com/aicers/giganto-client/commit/6fe76541ecc702165f92d9dd9849c79120568b46, public structs in `giganto-client` no longer depend on `jiff` or `chrono`, have removed `end_time` fields, and reverted from bincode v2 to bincode v1. Since this matches the intended usage of `giganto-client` by Giganto, Giganto will be updated accordingly.

With this update, the chrono-to-jiff migration scope in Giganto is minimized. Only the GraphQL API part of the external interface requires verification, while the verification in terms of ingest, publish, and database schema interfaces become unnecessary. This is related to issue #1322.

Updating `giganto-client` to `6fe7654` also addresses issues #1333 and #1329. This change effectively revisits the previous migration in #1172, which replaced various time fields from `i64` to `chrono::DateTime` or `jiff::Timestamp`. We should carry out a meaningful adjustment,  not a simple revert.

Additionally, since `giganto-client@6fe7654` uses bincode v1, I believe it is beneficial to revert to bincode v1, as described in issue #1320.

In summary, this issue collectively addresses the interconnected migration concerns across the following issues:  
- #1322  
- #1333  
- #1329  
- #1320  

Rather than resolving these individually, the plan is to address them together in a single pull request, grouping commits by logical units. This integrated approach is expected to be far more efficient and effective than individual approach in the current situation.

Issue #1333, opened 2025-11-12
Remove `end_time` field from all raw event structs


kimhanbeom:
## Background 
- Currently, all raw event structs in giganto-client contain an `end_time` field.
- However, the combination of `start_time` and `duration` is sufficient to represent the event’s time range, making `end_time` redundant.
- Therefore the `end_time` field should be removed from all structs.

## Task
- Once this change is applied to giganto-client, the corresponding updates should also be reflected in giganto.
- Similarly, `end_time` should also be removed from all external-facing structures (e.g., GraphQL response types).

Issue #1329, opened 2025-11-06
Separate internal and external timestamp formats for clarity and efficiency


kimhanbeom:
## Background
- https://github.com/aicers/giganto-client/issues/234 
- The giganto-client has updated several Network and Sysmon structs to use `i64` (epoch nanoseconds) for timestamp fields such as `start_time`, `end_time`, `creation_utc_time`, and `previous_creation_utc_time`. This change improves transmission and storage efficiency.

## Task
- Netwokr Protocol and Sysmon Struct Updates: The timestamp fields (`start_time`, `end_time`, `creation_utc_time`, and `previous_creation_utc_time`) in the Network and Sysmon structs provided by giganto-client have been changed to `i64`. Update each corresponding struct to correctly handle.
  - Note: As mentioned in [giganto-client issue](https://github.com/aicers/giganto-client/issues/234) (Proposal section), external-facing structures (e.g., GraphQL response types) should continue to use rich timestamp types such as `Timestamp` or `DateTime<Utc>` for clarity and API consistency.

Issue #1322, opened 2025-11-04
Ensure `Chrono` → `Jiff` migration preserved public behavior through programmatic tests


henry0715-dev:
### Background

Before migrating to the Jiff crate, we need to add missing regression tests for time-related logic that currently depends on the `Chrono` crate.
By adding these tests, we can ensure that the existing program behavior remains consistent and stable even after the migration to `Jiff`.

### Task

- Identify all modules and functions that use the `Chrono` crate for time or date handling.
- Add regression tests to verify expected behavior and output formats.
- Include tests for time conversions, formatting, parsing, and boundary conditions (e.g., min/max timestamps).
- Ensure tests run consistently across environments (UTC vs Local time).
- Validate that all Chrono-based behaviors are fully covered before migrating to Jiff.

PR #1336, opened 2025-11-17
Update giganto-client to `6fe7654`

sophie-cluml:
There are 4 major commits.

The first commit 83aff2fd66134f649134770e6de20e2b3dc28ab9 uses `i64` for time data and removes `end_time`.  It will:

Close: #1329
Close: #1333

The second commit b1f9256ca9382a14f0bfc6b6eb1dfd49c05c7828 brings bincode back to v1. It will:

Close: #1320

The third commit 7aa5923a602b3814e5d0e233ebd662dbe2cc3fba adds time format related integration tests for GraphQL APIs, which is a prep for the chrono-to-jiff migration. And the fourth commit 5a16ebd0f57a81f1de9bce5d594a90ede019c4ff adds tests for storage key builder functions, which also serves as a prep for the chrono-to-jiff migration. However, even aside from the chrono-to-jiff effort, this work is valuable on its own because it adds missing unit tests for the storage key builders. The third and fourth commits will:

Close: #1322

All in all, this PR will:

Close: #1335

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,19 +52,18 @@ extern crate proc_macro;
 /// #[derive(GraphQLQuery)]
 /// #[graphql(
 ///     schema_path = "src/graphql/client/schema/schema.graphql",
-///     query_path = "src/graphql/client/schema/conn_raw_events.graphql",
+///     query_path = "src/graphql/client/schema/example_raw_events.graphql",
 ///     response_derives = "Clone, Default, PartialEq"
 /// )]
-/// pub struct ConnRawEvents;
+/// pub struct ExampleRawEvents;
 ///
 /// #[derive(SimpleObject, ConvertGraphQLEdgesNode)]
-/// #[graphql_client_type(names = [conn_raw_events::ConnRawEventsConnRawEventsEdgesNode, network_raw_events::NetworkRawEventsNetworkRawEventsEdgesNodeOnConnRawEvent])]
-/// struct ConnRawEvent {
+/// #[graphql_client_type(names = [example_raw_events::ExampleRawEventsExampleRawEventsEdgesNode, network_raw_events::NetworkRawEventsNetworkRawEventsEdgesNodeOnExampleRawEvent])]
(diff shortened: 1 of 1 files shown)


kimhanbeom (changes requested):
1. It seems the current changelog could be revised to better highlight the changes relative to the previous release (v0.25.1).
    - Text to be removed
       - Changed `last_time` field to `end_time` for raw events.....
       - Changed `duration` field to `end_time` for Conn raw events in....
    - Text to be added
       - Removed the `last_time` field from all network raw event structs. The combination of `start_time` and `duration` sufficiently represents the event’s time range, making `last_time` redundant.`
    - Text to be modified (replace `end_time` to `last_time`)
       - The `start_time` field represents the session start time and complements
        the existing `end_time` field.
2. In the src/lib.rs derive usage examples, could you please also remove the `end_time` field and update the type of the `start_time` field.

sophie-cluml:
@kimhanbeom Thank you for the review.

- I removed and added the CHANGELOG as you suggested. However, regarding your suggestion on the modification, I chose to remove the phrase ```and complements the existing `end_time` field``` instead. I believe ```The `start_time` field represents the session start time``` already explains the meaning of `start_time` sufficiently on its own.
- Regarding your comment on `lib.rs`: I don’t think the `ConnRawEvent` struct in the documentation example must match  the struct that Giganto defines in production code. The purpose of the example is to provide helpful reference information for anyone looking to understand `lib.rs` (mainly with various field types). Since we already have an `i64` field in the example, I believe removing `end_time` makes the example simple and concise. I removed it, and since I believe this is unrelated to the actual removal of the `end_time` field in production code, I separated the removal into its own  commit f3de6eb. And to avoid any misunderstanding that the documentation example struct must match the production struct, I also adjusted the example struct name.

NOTE: With the latest force-push, only the CHANGELOG of the first commit was adjusted and the last commit was newly added.

zmrdltl (approved):
Looks good—revert back to bincode v1 is clean and consistent

PR #1397, opened 2025-12-30
Document essential tools for build

### LINKED ISSUE #1396: Update build documentation for Ubuntu dependencies
## Description

Currently, building giganto on Ubuntu may fail due to missing system dependencies. 
The current documentation lacks specific instructions on the necessary build tools, which can be confusing for contributors.

## Root Cause

The build failures are primarily caused by the absence of essential compilation tools.

### Missing C Compiler/Linker (cc)

Crates like `quote` and `serde` require a C compiler to execute build scripts. Without it, the following error occurs:

Image

### Missing Clang/LLVM

The `rocksdb` crate requires clang to generate bindings via bindgen. If missing, the build fails with:

Image

## Proposed Solution

I suggest adding a "Prerequisites" or "Build Instructions" section to the `README.md` to guide users through installing the required dependencies.

For Ubuntu users, the following packages should be installed before building giganto:

```sh
sudo apt update
sudo apt install build-essential clang llvm-dev libclang-dev
```

JonghoKim-jj:
Close #1396

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,13 @@ and real-time analytics.
 - High Performance: Giganto is designed to efficiently handle high volumes of
   data, ensuring optimal performance for AICE.
 
+## Build
+
+### Build Requirements
+
+- libpcap
+- Clang & LLVM: To build [rocksdb](https://crates.io/crates/rocksdb)
+
 ## Usage
 
 You can run Giganto by invoking the following command:


sophie-cluml (commented):
I left a comment here: https://github.com/aicers/giganto/issues/1396#issuecomment-3709225872

PR #1421, opened 2026-01-13
Log skipped entries during `OpLog` migration

### LINKED ISSUE #1420: Improve migration robustness and error observability
## Background

The current migration logic lacks sufficient defensive handling and observability for certain edge cases, which can negatively impact data reliability and make root-cause analysis difficult in production environments.

The following issues have been identified:

- During HTTP migration, negative duration values can occur when `end_time` is earlier than `start_time`, which may happen due to clock skew between systems.
- In OpLog migration, skipped entries are not sufficiently reported, making it difficult to monitor migration quality or debug data inconsistencies.

## Goal

- Make migration logic robust against invalid or inconsistent input
- Improve visibility and observability of migration errors and skipped data

## Tasks

- HTTP migration
  - lamp duration to zero when `end_time` < `start_time`
  - Ensure graceful handling of clock skew scenarios without panics or invalid values
- OpLog migration
  - Track skipped entries during migration
  - Emit detailed warning logs including skip reasons
  - Enhance logging to better assess overall migration quality

## Expected Outcome

- Stable migration behavior even under time inconsistency scenarios
- Clear insight into skipped data and migration issues
- Improved operational monitoring and debugging of migration processes

henry0715-dev:
Close #1420

### DIFF
--- a/src/storage/migration.rs
+++ b/src/storage/migration.rs
@@ -191,9 +191,11 @@ fn migrate_0_23_0_to_0_24_0_op_log(db: &Database) -> Result<()> {
     info!("Starting migration for oplog");
     let store = db.op_log_store()?;
     let counter = AtomicUsize::new(0);
+    let mut skipped_count = 0;
 
     for raw_event in store.iter_forward() {
         let Ok((key, value)) = raw_event else {
+            skipped_count += 1;
             continue;
         };
 
@@ -201,6 +203,7 @@ fn migrate_0_23_0_to_0_24_0_op_log(db: &Database) -> Result<()> {
             get_timestamp_from_key(&key),
             bincode::deserialize::<OpLogBeforeV24>(&value),
         ) else {
+            skipped_count += 1;
             continue;
         };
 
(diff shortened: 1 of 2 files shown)


src/storage/migration.rs
```diff
@@ -191,16 +191,21 @@ fn migrate_0_23_0_to_0_24_0_op_log(db: &Database) -> Result<()> {
     info!("Starting migration for oplog");
     let store = db.op_log_store()?;
     let counter = AtomicUsize::new(0);
+    let mut skipped_count = 0;
```
sophie-cluml: Migration functions like `migrate_0_24_to_0_26_conn` and `migrate_0_24_to_0_26_http` return errors when rows fail to be migrated, and this naturally guarantees observability. However, we are taking a different approach for oplogs because they are solution module logs, not event data. For oplogs, it is intended to drop problematic rows. To balance this with observability, I suggest a compromise: log a single summary line instead of detailed errors for each row. Also, since the drops are intended I think logging as `info` level suits our needs.

src/storage/migration.rs
```diff
@@ -386,7 +396,7 @@ fn migrate_0_24_to_0_26_http(db: &Database) -> Result<()> {
             resp_port: old.resp_port,
             proto: old.proto,
             start_time,
-            duration: old.end_time - start_time,
+            duration: (old.end_time - start_time).max(0),
```
sophie-cluml: Applying `max(0)` correction only to the HTTP `duration` field creates an inconsistency with other protocols, which simply maintain the `end_time - start_time` pattern. Most importantly, the motivation for `migrate_0_24_to_0_26_http` is to apply a structural transformation to match the new schema, not to correct data. Actively altering data values exceeds the scope of this migration and should be avoided.

sophie-cluml (changes requested):

PR #1438, opened 2026-01-19
Add GraphQL tests for Export operations

### LINKED ISSUE #1426: Add GraphQL tests for Export operations
Add tests for representative Export queries/mutations. Follow parent issue requirements.

dayeon5470:
Close #1426

### DIFF
--- a/src/graphql/export/tests.rs
+++ b/src/graphql/export/tests.rs
@@ -1,53 +1,150 @@
-use std::fmt::Write as _;
-use std::net::IpAddr;
+use std::fs;
+use std::net::SocketAddr;
 use std::sync::{Arc, OnceLock};
-use std::{fs, io::ErrorKind, mem, path::PathBuf, str::FromStr};
 
-use chrono::{Duration as ChronoDuration, TimeZone, Utc};
-use giganto_client::{
-    RawEventKind,
-    ingest::{
-        log::{Log, OpLog, OpLogLevel, SecuLog},
-        netflow::{Netflow5, Netflow9},
-        network::{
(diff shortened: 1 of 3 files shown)


src/graphql/export/tests.rs
```diff
@@ -86,6 +201,148 @@ async fn invalid_query() {
...
+            filter:{
+                protocol: "conn",
+                sensorId: "src1",
+                agentName: "agent",
+                agentId: "agent_id"
+            }
+            ,exportType:"json")
+    }"#;
+    let res = schema.execute(query).await;
+    assert_eq!(res.data.to_string(), "null");
```
henry0715-dev: In the `invalid_query` test, the assertion only checks `assert_eq!(res.data.to_string(), "null")` rather than validating the actual error.
Since the error details are available in `res.errors`, it would be better to assert on the specific error content to clearly verify which error is being returned.

henry0715-dev:
`export_file_warns_on_invalid_data` and `export_oplog_file_warns_on_invalid_data` mention “warns” in their names, but they do not actually assert any warning logs. Instead, they only check that the functions return `Ok` and that the output files exist. The test names and their assertions are therefore inconsistent.

dayeon5470:
> `export_file_warns_on_invalid_data` and `export_oplog_file_warns_on_invalid_data` mention “warns” in their names, but they do not actually assert any warning logs. Instead, they only check that the functions return `Ok` and that the output files exist. The test names and their assertions are therefore inconsistent.

I’ve made all the changes according to your feedback.

henry0715-dev:
The `tests.rs` file mixes a lot of helper/fixture-style functions alongside the test bodies, which breaks the reading flow. For readability, it would be better to move those helpers into a dedicated `fixture` module (e.g., `mod fixture;` or `tests/fixture.rs`) and keep `tests.rs` focused on the test cases.

PR #1452, opened 2026-01-19
Add GraphQL tests for Sysmon operations

### LINKED ISSUE #1434: Add GraphQL tests for Sysmon operations
Add tests for representative Sysmon queries/mutations. Follow parent issue requirements.

dayeon5470:
Close #1434


henry0715-dev (approved):

henry0715-dev:
Insufficient negative coverage in Search tests
There are no tests verifying that searches return empty results when filters do not match, or that filtering is correctly applied based on specific fields (e.g., `queryName`, `hashes`, etc.).

henry0715-dev:
Current tests focus on success/empty results and do not validate GraphQL error responses for invalid queries (e.g., invalid cursor, using both `after` and `before`). Please add failure-query cases to cover error handling.

henry0715-dev:
In `sysmon_events_local_cases` and `sysmon_events_cluster_cases`, only the “same cursor” error path covers a non-success scenario. Other invalid combinations—such as `after + last`, `before + first`, or out-of-range cursors—are currently not covered.

It would be better to add explicit test cases for these failure combinations to ensure pagination validation logic is thoroughly verified and to prevent regression in edge-case handling.

henry0715-dev:
@dayeon5470 
It might be better to split the helper functions out into a dedicated `fixture` module to keep the tests easier to read and extend.

sophie-cluml (approved):

PR #1453, opened 2026-01-19
Add tests for graphql.rs

### LINKED ISSUE #1435: Add tests for graphql.rs
Add integration-style tests for the root GraphQL schema wiring (schema setup, context/auth injection, error mapping). Follow parent issue requirements.

dayeon5470:
Close #1435

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -917,20 +917,28 @@ impl_string_number!(StringNumberI64, i64);
 
 #[cfg(test)]
 mod tests {
-    use std::collections::{HashMap, HashSet};
+    use std::collections::{BTreeSet, HashMap, HashSet};
+    use std::net::IpAddr;
     use std::sync::Arc;
 
     use async_graphql::EmptySubscription;
+    use chrono::{DateTime, TimeZone, Utc};
+    use serde::{Deserialize, Serialize};
     use tokio::sync::Notify;
 
     use super::{
-        NodeName, StringNumberI64, StringNumberU32, StringNumberU64, StringNumberUsize, schema,
+        NodeName, Result, SearchFilter, StringNumberI64, StringNumberU32, StringNumberU64,
+        StringNumberUsize, TIMESTAMP_SIZE, TimeRange, check_address, check_agent_id,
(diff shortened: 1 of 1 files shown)


src/graphql.rs
```diff
@@ -810,7 +810,8 @@ fn check_port(filter_port: Option<&PortRange>, target_port: Option<u16>) -> bool
             let ends_before = port_range.end.is_none_or(|end| port < end);
             starts_after_or_at && ends_before
         }
-        _ => true,
+        (Some(_), None) => false,
+        (None, _) => true,
```
sophie-cluml: @dayeon5470 @henry0715-dev Is this code change intended? Is this a bug fix?

sophie-cluml (changes requested):

src/graphql.rs
henry0715-dev: I missed the production code modification during PR.
Thank you for checking.

src/graphql.rs
```diff
@@ -1158,4 +1207,261 @@ mod tests {
...
+    #[test]
+    fn test_check_port() {
+        let filter = PortRange {
+            start: Some(1000),
+            end: Some(1002),
+        };
+
+        assert!(check_port(Some(&filter), Some(1000)));
+        assert!(!check_port(Some(&filter), Some(1002)));
+        assert!(check_port(Some(&filter), None));
```
sophie-cluml: Tests that assert incorrect behavior are worse than having no tests. Please don’t add assertions for behavior we believe is wrong—if it’s incorrect, the test shouldn’t validate it. This isn’t specific to this PR: if we find a test validating incorrect behavior, we should treat that test as a bug and fix it. Adding `assert!(check_port(Some(&filter), None));` in this PR is effectively introducing a known-bad assertion.

In the same spirit, if there are any similar asserts that should be removed, please remove them in this PR. If that isn’t feasible right now, adding a clear TODO (ideally with an issue number) might be an alternative.

dayeon5470: I’ve removed the test assertion that was validating the incorrect behavior.

PR #1462, opened 2026-01-20
Add tests for ingest/tests.rs

### LINKED ISSUE #1444: Add tests for `ingest/tests.rs`
EOM

henry0715-dev:
Close #1444

### DIFF
--- a/src/comm/ingest/tests.rs
+++ b/src/comm/ingest/tests.rs
@@ -2,17 +2,30 @@
 
 use std::{
     fs,
+    future::IntoFuture,
     net::{IpAddr, Ipv6Addr, SocketAddr},
     path::Path,
-    sync::{Arc, OnceLock},
+    sync::{
+        Arc, OnceLock,
+        atomic::{AtomicI64, Ordering},
+    },
 };
 
 use base64::{Engine, engine::general_purpose::STANDARD as base64_engine};
 use chrono::{Duration, TimeZone, Utc};
+use giganto_client::frame::SendError;
+use giganto_client::ingest::log::SecuLog;
+use giganto_client::ingest::netflow::{Netflow5, Netflow9};
+use giganto_client::ingest::network::{MalformedDns, Radius};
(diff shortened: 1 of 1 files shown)


src/comm/ingest/tests.rs
```diff
@@ -1363,36 +1534,270 @@ async fn one_short_reproduce_channel_close() {
...
+
+    super::send_ack_timestamp(&mut send, 100).await.unwrap();
+
+    // Now FINISH/CLOSE the stream to force a failure on next send?
+    send.finish().unwrap(); // Half-closed
+
+    // Sending on a finished stream should fail?
+    // Quic SendStream: "Writing to a stream that has been finished or reset will return an error."
+    let res = super::send_ack_timestamp(&mut send, 200).await;
+    assert!(res.is_err());
```
dayeon5470: Instead of just checking is_err(), it would be better to assert on the specific error to ensure the failure is for the expected reason.

dayeon5470 (dismissed):

src/comm/ingest/tests.rs
```diff
@@ -152,1247 +305,1974 @@ fn init_client() -> Endpoint {
...
-    send_record_header(&mut send_op_log, RAW_EVENT_KIND_OPLOG)
+    send_record_header(&mut send_log, RAW_EVENT_KIND_LOG)
         .await
         .unwrap();
-    send_events(
-        &mut send_op_log,
-        Utc::now().timestamp_nanos_opt().unwrap(),
-        op_log_body,
+    let base_timestamp = next_timestamp();
+    send_events(&mut send_log, next_timestamp(), log_body)
```
dayeon5470: ack_interval_sends_last_timestamp uses next_timestamp() for the first event, making it base_timestamp + 1, which is duplicated by the first loop iteration. This likely doesn’t affect the ACK, but if the goal is a clean monotonic 1024-event sequence, send the first event with base_timestamp instead.

PR #1463, opened 2026-01-20
Add tests for publish/tests.rs

### LINKED ISSUE #1446: Add tests for `publish/tests.rs`
EOM

henry0715-dev:
Close #1446


dayeon5470:
`process_pcap_extract_filters` must continue processing remaining filters/peers even if one fails (partial failure is allowed).
Add a test that verifies the second filter is still executed after the first filter fails.

henry0715-dev:
I’ve updated everything.

PR #1498, opened 2026-02-13
Fix check_address/check_port reject None target

### LINKED ISSUE #1497: Fix `check_address` / `check_port` to return false for (Some, None)
Description
Currently, `check_address` and `check_port` only apply filtering for (Some(filter), Some(target)). All other cases return true, so (Some(filter), None) incorrectly passes.

Change
Update both functions so that when a filter is provided but the target is missing, it returns false:

(None, _) -> true

(Some(_), None) -> false

(Some(_), Some(_)) -> range check

dayeon5470:
Close #1497


kimhanbeom (changes requested):
@octoaide 
- Could you please rebase the PR.
- Move the current changes content under the `Fixed` section of `Unreleased`.

Issue #1535, opened 2026-03-10
Introduce MkDocs Multilingual Documentation Structure


jinjoolee07:
## Background
The project currently has no MkDocs-based documentation setup, so we will introduce a minimal baseline (EN/KO, strict build, CI validation, and Pages deployment).

## Goals
- Add documentation build configuration based on `mkdocs.yml`
- Establish multilingual docs structure under `docs/en` and `docs/ko`
- Use the same MkDocs dependencies in both local and CI environments
- Enforce `mkdocs build --strict` as a required CI validation step
- Add a GitHub Pages deployment workflow if needed

## Tasks
1. Create `mkdocs.yml`
2. Create `docs/en` and `docs/ko` folders, with matching document filenames in both locales
3. Finalize MkDocs dependency installation path
   - Local: document virtualenv-based (`.venv`) installation
   - CI: add workflow steps to install `mkdocs-material` and `mkdocs-static-i18n`
4. Add a `mkdocs build --strict` step to CI
5. Add `docs.yml` GitHub Pages deployment workflow if needed

Issue #1526, opened 2026-03-05
Create EN/KR user manuals for Giganto


dayeon5470:
## Goal
Deliver a complete, production-grade user/operator manual (EN/KR) for Giganto so that a non-expert can install, configure, run, and operate it in both standalone and cluster modes without relying on external references. Publish PDF output as part of the docs deliverables.

## Scope
- Full manual content in English and Korean
- Standalone installation and operation guide
- Cluster deployment and operation guide
- GraphQL query and API usage guide
- PDF output from the documentation site

PR #1544, opened 2026-03-17
Add Giganto manual and initialize MkDocs

dayeon5470:
Close: #1535
Close: #1526


jinjoolee07:
1. Reduced duplication in `graphql.md` and kept only operational guidance such as mTLS and Playground access.
2. Added Rustdoc comments in the GraphQL API so the generated schema / Playground can serve as the primary reference.
3. Left a comment in the issue that the current `docs/` and `scripts/` area may be a future extraction target: https://github.com/aicers/giganto/issues/1535#issuecomment-4087862609

sophie-cluml (commented):
I noticed that documents under docs/ko use very short line wrapping. Is there a particular reason for this? It’s not a Markdown issue, but the frequent line breaks seem to reduce readability. Was this intentional, or possibly caused by editor settings? If unintentional, could you reflow the text please?

dayeon5470:
Thanks for pointing that out. I’ve reflowed the text and updated the formatting.

dayeon5470:
@sophie-cluml Reviewed against the guidance in [aice-web-next#178](https://github.com/aicers/aice-web-next/discussions/178).

mkdocs build --strict, and the i18n and PDF build flows are working as expected. I do see the Material for MkDocs warning about MkDocs 2.0 during both the docs build and PDF generation, but it appears to be an upstream warning and is not affecting the current setup.

Regarding the GitHub Actions workflow, you previously advised not to include it in this PR. Based on the current status, should we continue to leave it out, or is there now a need to add it?

sophie-cluml:
> This PR is meant as an initial documentation skeleton rather than a full manual.

I’d appreciate it if this context could also be captured in the linked issue, so that later, even without digging into the PR implementation details or comments, both humans and AI can still understand the work from a high-level, overall perspective.

PR #1732, opened 2026-08-04
Migrate from deluxe to darling and bump syn to 3

### LINKED ISSUE #1731: Migrate from deluxe to darling and bump syn to 3
# Migrate from deluxe to darling and bump syn to 3

## Background

Dependabot PR #1718 bumps `syn` from `2.0.119` to `3.0.3`. However, `giganto` currently uses `deluxe 0.5` for proc-macro attribute parsing, and `deluxe 0.5` depends on `syn 2`.

As a result, the `syn 3` types used directly by `giganto` and the `syn 2` types expected by `deluxe` are treated as different types, which causes trait bound errors. Therefore, the `syn` bump cannot be merged as a standalone dependency update.

The failures are in the following areas:

- `Vec<syn::Type>: ParseMetaItem` trait bound mismatch
- `syn::Error` cannot be converted into `deluxe::Error`
- `syn::DeriveInput: deluxe::HasAttributes` trait bound mismatch
- `syn::Field: deluxe::HasAttributes` trait bound mismatch

## Proposal

Instead of keeping `deluxe` and trying to bump `syn` to 3, migrate the proc-macro attribute parsing to `darling 0.24`. `darling 0.24` uses `syn 3`, so `giganto` can bump its direct `syn` dependency to 3 at the same time.

This is a code migration, not just a dependency bump. It should be handled in a human-owned follow-up PR rather than by adding commits to the Dependabot PR #1718. PR #1718 should be superseded by this issue and the follow-up PR.

## Scope

- Remove the `deluxe = "0.5"` dependency.
- Add the `darling = "0.24"` dependency.
- Bump the direct `syn` dependency to `3`.
- Migrate the `ConvertGraphQLEdgesNode` derive macro attribute parsing from `deluxe` to `darling`.
- Preserve the existing `#[graphql_client_type(...)]` attribute syntax.
- In particular, `names = [module::Type, ...]` must continue to work as before.
- Verify that `deluxe` is removed from the dependency graph.

## Prior Check

A local spike confirmed that, after removing `deluxe` and applying `darling 0.24` with `syn 3.0.3`, the project can build and start with the `cluster` feature enabled.

Checked:

- `cargo fmt --check` passed
- `cargo check --locked --no-default-features --features cluster` passed
- The binary started successfully with the `cluster` feature enabled
- The GraphQL endpoint responded when using test certificates

Full CI and schema validation still need to be checked in the PR that resolves this issue.

Implementation notes:

- The existing `#[graphql_client_type(names = [module::Type, ...])]` syntax must be preserved.
- `darling 0.24` can handle `syn::Type`, but the current array-shaped attribute may require additional parsing logic to preserve the existing syntax.

## Acceptance Criteria

- `deluxe` is removed from `Cargo.toml`.
- `darling 0.24` is added to `Cargo.toml`.
- The direct `syn` dependency in `Cargo.toml` is bumped to `3`.
- The existing `#[graphql_client_type(...)]` attribute syntax remains compatible.
- `cargo fmt --check` passes.
- `cargo check --locked --no-default-features --features cluster` passes.
- Schema-related CI passes.
- The server starts with the `cluster` feature enabled and the GraphQL endpoint responds.

sophie-cluml:
## Summary

Closes #1731.
Supersedes #1718.

This migrates the `ConvertGraphQLEdgesNode` proc-macro attribute parsing from `deluxe` to `darling` so that `giganto` can move its direct `syn` dependency to 3.

Changes:
- Replace `deluxe = "0.5"` with `darling = "0.24"`.
- Bump the direct `syn` dependency from 2 to 3.
- Preserve the existing `#[graphql_client_type(...)]` syntax, including `names = [module::Type, ...]`.
- Add focused unit coverage for the new attribute parsing paths.
- Remove `deluxe` from the dependency graph.

## Verification

- `cargo fmt --check`
- `cargo clippy --locked --no-default-features --features cluster -- -D warnings`
- `cargo clippy --features bootroot --tests -- -D warnings`
- `cargo check --locked --no-default-features --features cluster`
- `cargo check --locked --all-features`
- `cargo test -q`
- `cargo test -q schema_should_be_up_to_date -- --ignored`
- Started the binary with `cluster` enabled using test certificates and confirmed the GraphQL endpoint responds.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,11 +2,12 @@
 
 use core::panic;
 
+use darling::{FromDeriveInput, FromField, FromMeta};
 use proc_macro::TokenStream;
 use quote::{ToTokens, quote};
 use syn::{
-    Data, DataStruct, DeriveInput, Fields, GenericArgument, Ident, PathArguments, Type, parse_str,
-    parse2,
+    Data, DataStruct, DeriveInput, Expr, Fields, GenericArgument, Ident, PathArguments, Type,
+    parse_str, parse2,
 };
 
 extern crate proc_macro;
@@ -99,8 +100,8 @@ extern crate proc_macro;
 ///             orig_port: node.orig_port.map(|x| x as _),
 ///             proto: node.proto as _,
 ///             start_time: node.start_time as _,
-///             service: node.service as _,
-///             resp_pkts: node.resp_pkts as _,
+///             service: node.service,
+///             resp_pkts: node.resp_pkts,
 ///             ttl: node.ttl.into_iter().map(|x| x as _).collect(),
 ///             filenames: node.filenames,
 ///             ja3s: node.ja3_s,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -99,8 +100,8 @@ extern crate proc_macro;
 ///             orig_port: node.orig_port.map(|x| x as _),
 ///             proto: node.proto as _,
 ///             start_time: node.start_time as _,
```
kimhanbeom: @octoaide
In the current PR, the `as _` expressions in the documentation examples have only been partially updated. Since the macro generates `as _` only for the types listed in `SAFE_TO_I64_TYPES` (`u8`, `u16`, `i8`, `i16`, and `i32`), please remove `as _` from `start_time`, `service`, and `resp_pkts` in both `From` examples so that the documentation matches the actual generated code.

kimhanbeom (changes requested):

kimhanbeom (approved):

